PR #31, opened 2020-11-17
Add Scss pattern

adamjarling:
This is a partial PR, but I know others are working on potential styling UI so wanted to get this in as an example of how we can structure styling in the app.

References this color chart:
https://www.schemecolor.com/20-best-gray-colors-for-ui-web-design.php

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "/dist"
   ],
   "scripts": {
-    "build": "NODE_ENV=production rollup -c",
+    "build": "NODE_ENV=production rollup -c && styleguidist build",
     "prepublishOnly": "npm run build",
     "dev": "styleguidist server",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
     "/dist"
   ],
   "scripts": {
-    "build": "NODE_ENV=production rollup -c",
+    "build": "NODE_ENV=production rollup -c && styleguidist build",
```
divyareddynu, line 12: I remember this was removed in your previous PR, are keeping it?

adamjarling: Yea, that was an accident on my end.  Including that last statement just re-builds the Styleguidist static site documentation which gets pushed to the Github repo `/docs` folder.

divyareddynu (approved):
Loads up nicely!

PR #38, opened 2020-11-19
Removes the duplicate instantiations of player caused by startTime en…

adamjarling:
…dTime useEffect()

This removes the duplicate instantiations of `Player` which were getting called on every `useEffect()` render on `startTime` and `endTime` switching.   Now we're just dealing with one instance of the player, attached to the `PlayerProvider`

This might impact your work @Dananji ....fyi.  The new Player everywhere.


phuongdh (approved):
Thanks for fixing this. I wasn't sure why we didn't get the player from playerState.

PR #42, opened 2020-11-23
Playlist firstpass

divyareddynu:


src/components/Playlist/Playlist.js
```diff
@@ -0,0 +1,70 @@
...
+    fetch(manifestUrl)
+      .then((result) => result.json())
+      .then((data) => {
+        dispatch({ manifest: data, type: 'updateManifest' });
+        setActiveItem(manifestUrl);
+        console.log(data);
+
+        return data;
+      });
+  };
```
adamjarling, line 46: I'd say for now we just hardcode some mock display data.  Not sure that we'll want to make an extra network request for each manifest in a Playlist manifest.  There's probably a better way to serve this data up for the front-end.

adamjarling (changes requested):

PR #44, opened 2020-11-23
Structure nav

Dananji:
This PR contains the following;
1. Handle source transition between same source types (audio/video) for `ended` event from the VideoJS player
2. Structure navigation across same source types (audio/video)

### DIFF
--- a/src/components/MediaPlayer/VideoJSPlayer.js
+++ b/src/components/MediaPlayer/VideoJSPlayer.js
@@ -41,14 +41,12 @@ function VideoJSPlayer({
       ...videoJSOptions,
     };
 
+    setCIndex(canvasIndex);
+
     const newPlayer = videojs(playerRef.current, options);
 
     newPlayer.getChild('controlBar').addChild('vjsYo', {});
 
-    if (isClicked && canvasIndex !== cIndex) {
-      const oldPlayer = videojs(`videojs-player-${canvasIndex}`);
-      switchPlayer(oldPlayer);
-    }
     playerDispatch({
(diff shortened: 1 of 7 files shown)


src/components/MediaPlayer/VideoJSPlayer.js
```diff
@@ -89,44 +111,38 @@ function VideoJSPlayer({ isVideo, initStartTime, ...videoJSOptions }) {
...
-    //     captionOn,
-    //     manifest
-    //   );
-    //   playerDispatch({ player: newInstance, type: 'updatePlayer' });
-    //   setCIndex(cIndex + 1);
-    // }
+  const handleEnded = () => {
+    if (hasNextSection({ canvasIndex, manifest })) {
+      manifestDispatch({ canvasIndex: canvasIndex + 1, type: 'switchCanvas' });
+      const oldPlayer = videojs(`videojs-player-${cIndex}`);
```
adamjarling, line 128: @Dananji Does this create a new instance of a player?  Let's chat on this tomorrow.  I could be wrong, but seems like making multiple calls to `videojs()` is going to create a bunch of different scopes of players?

Dananji: According to the docs [here](https://docs.videojs.com/tutorial-player-workflows.html) it seems to only access the existing players. But as you said there could be side effects from this call.

adamjarling: Ah, interesting.

adamjarling (approved):

PR #49, opened 2020-11-25
Update markers styling

### LINKED ISSUE #48: Update markers UI
### Description
Update markers UI in VideoJS player.

### Done looks like
- [ ] Marker color changed from red to Avalon green
- [ ] Marker highlight size remains contained and doesn't overlap regions.

adamjarling:
closes #48 

image

### DIFF
--- a/src/components/MediaPlayer/VideoJSPlayer.js
+++ b/src/components/MediaPlayer/VideoJSPlayer.js
@@ -83,9 +83,11 @@ function VideoJSPlayer({
             },
           },
           markerStyle: {
-            width: '4px',
-            'background-color': 'red',
+            opacity: '0.5',
+            'background-color': '#80A590',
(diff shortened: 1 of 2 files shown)


phuongdh:
@adamjarling If we use the same library for playlist markers and section highlights then this styling may not work for both.

adamjarling:
@phuongdh Yea you may be right.  I'm thinking let's just push through a `v1` release for today that isn't concerned with Playlists quite yet.  Updating this styling seems pretty straight forward and we can pivot quickly on that later?  

A better solution might be eventually developing our own videojs plugin for markers.  Looks like this markers package is no longer maintained.   But we might be able to see how they wired it up, then customize for how we need it?   Or we start helping maintain this package or something?

Dananji (approved):
:+1:

PR #51, opened 2020-11-25
Bug fixes when switching the player

Dananji:
This PR includes;
1. Bug fix for when switching audio -> video player with play() event
2. Marking the current nav item when advancing from canvas to next when the current media ends playing


adamjarling (approved):
Works great.

PR #61, opened 2021-01-25
Add instructions to static demo page for using external manifest URLs

Dananji:
Instructions in the comments block in the code editor:

Screenshot from 2021-01-25 12-54-15


adamjarling (approved):
Looks good to me.

PR #62, opened 2021-02-25
Improve timerail highlight

Dananji:
Player highlighting the structure in the time rail with structure navigation and seek/scrub events:

https://user-images.githubusercontent.com/1331659/109183291-c83b7080-775b-11eb-893f-5c725d08044d.mp4

Timerail highlight updating with `timeupdate` event from `video.js`:
(First the existing timerail highlight is removed when the time reaches the end of that media fragment (around 06:13). From 06:13-08:45 there is no defined structure, so there's no timerail highlight. And then when time reaches 08:45 the highlight is created for the media fragment starting from 08:45 timestamp)  

https://user-images.githubusercontent.com/1331659/110801233-00a67880-824b-11eb-955d-ebadbf2aad96.mp4


src/components/MediaPlayer/VideoJSPlayer.js
```diff
@@ -116,66 +141,197 @@ function VideoJSPlayer({
...
   React.useEffect(() => {
     if (!player || !currentPlayer) {
       return;
     }
-
-    if (startTime != null) {
-      player.currentTime(startTime, playerDispatch({ type: 'resetClick' }));
-
-      // Mark current timefragment
+    if (currentNavItem != null && isReady) {
```
adamjarling, line 175: Is there a reason to not use `currentNavItem !== null` here?   Just a heads up coercion could possibly bring up some odd behavior...maybe?

src/components/MediaPlayer/VideoJSPlayer.js
```diff
@@ -116,66 +141,197 @@ function VideoJSPlayer({
...
           {
-            time: startTime,
-            duration: endTime - startTime,
+            time: start,
+            duration: stop - start,
             text: currentNavItem.label.en[0],
           },
         ]);
       }
+    } else if (startTime == null) {
```
adamjarling, line 194: Same, here on the `==` vs `===` just to be more explicit unless there's a reason not to?

src/components/MediaPlayer/VideoJSPlayer.js
```diff
@@ -116,66 +141,197 @@ function VideoJSPlayer({
...
+      // When canvas gets loaded into the player, set the currentNavItem and startTime
+      // if there's a media fragment starting from time 0.0.
+      // This then triggers the creation of a fragment highlight in the player's timerail
+      const firstItem = getSegmentMap({ manifest, canvasIndex })[0];
+      const timeFragment = getMediaFragment(getItemId(firstItem));
+      if (timeFragment && timeFragment.start == 0) {
```
adamjarling, line 200: Same here...

src/components/MediaPlayer/VideoJSPlayer.js
```diff
@@ -116,66 +141,197 @@ function VideoJSPlayer({
...
+        manifestDispatch({ item: null, type: 'switchItem' });
+      }
 
       handleIsEnded();
 
       setCIndex(cIndex + 1);
     }
   };
 
+  /**
```
adamjarling, line 286: These comments are great.  Nice work

adamjarling (approved):
All in all looks great.  Nice work.

src/components/MediaPlayer/VideoJSPlayer.js
Dananji: Yes, I agree. It's safe to use strict equality in place of these. Especially in cases where, we are comparing objects. Good catch! I will change these.

src/components/MediaPlayer/VideoJSPlayer.js
Dananji: Yeah, I thought it would be helpful to anyone contributing to the code in the future (including me :D). We didn't have much time to add comments when we were working on the community sprint, so I had to spent some time on understanding what the code is doing in a couple places.

PR #68, opened 2021-04-02
Add quality selector

Dananji:
quality-selector


adamjarling (approved):
Looks great and fairly easy to implement eh?   Nice work.

Dananji:
Yes, the integration is easy and very clean once I was able to figure out what code to use :D I think the documentation on the plugin could be more clearer (or it could just be me :facepalm: )

PR #86, opened 2021-07-21
Transcript UI implementation

Dananji:
This is how the transcript component looks with the changes in this PR:

transcript


src/components/Transcript/Transcript.js
```diff
@@ -0,0 +1,113 @@
...
+  const [transcript, setTranscript] = React.useState([]);
+  const [transcriptTitle, setTranscriptTitle] = React.useState('');
+  const [transcriptUrl, setTranscriptUrl] = React.useState('');
+
+  // React refs array for each timed text value in the transcript
+  let textRefs = React.useRef([]);
+  const transcriptContainerRef = React.useRef();
+
+  React.useEffect(() => {
+    if (transcripts && transcripts.length > 0) {
```
adamjarling, line 16: Might be able to short-circuit this evaluation with a `if (transcripts?.length > 0)...`

src/components/Transcript/Transcript.js
```diff
@@ -0,0 +1,113 @@
...
+      autoScrollAndHighlight(73004);
+    }, 3000);
+    setTimeout(function () {
+      autoScrollAndHighlight(27531);
+    }, 6000);
+  };
+
+  const autoScrollAndHighlight = (currenttime) => {
+    let textTopOffset = 0;
+    textRefs.current.map((tr) => {
```
adamjarling, line 51: `map` returns an array, so if you just looking to loop through `textRefs`, might be more clear to do a `forEach` or a `for...of` or `for...in`?

src/components/Transcript/TranscriptMenu/TranscriptSelector.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import TranscriptDownloader from './TranscriptDownloader';
+
```
adamjarling, line 3: Might want to pull in the `prop-types` package just to make it more clear to other devs, which props are expected for the component?

adamjarling (approved):

src/components/Transcript/Transcript.js
Dananji: Didn't know that javascript had that :100:

src/components/Transcript/TranscriptMenu/TranscriptSelector.js
Dananji: Absolutely!

PR #97, opened 2021-09-30
Use annotations instead of rendering in parser

Dananji:
Use annotations with supplementing motivation instead of rendering property in the transcript parser when extracting transcript data into the component.


src/components/Transcript/Transcript.md
```diff
@@ -54,25 +54,25 @@ import Transcript from '../Transcript/Transcript';
...
-          // Directly feeding a plain text file from a server
-          title: 'Transcript in Plain Text',
-          url: 'https://dlib.indiana.edu/iiif_av/iiif-player-samples/transcripts/transcript_plain.txt',
-        },
-        {
-          // External plain text transcript fed through `rendering` prop in a IIIF manifest
+          // External plain text transcript fed through `annotations` prop in a IIIF manifest
           title: 'External text transcript',
-          url: 'https://dlib.indiana.edu/iiif_av/iiif-player-samples/transcripts/transcript-manifest-rendering.json', // URL of the manifest
+          url: 'https://dlib.indiana.edu/iiif_av/iiif-player-samples/transcripts/transcript-manifest.json', // URL of the manifest
```
cjcolvar, line 59: I see that the json fixture filename is renamed here but it looks like the actual file needs to be renamed to match this change.  Same for `transcript-canvas-rendering.json` -> `transcript-canvas.json` below.

cjcolvar (approved):
This looks good to me.  I only had one question about filenames but it shouldn't blocking merging.

src/components/Transcript/Transcript.md
Dananji: Thanks for the review @cjcolvar.
Do you mean the file in the `test_data` directory in the code?

cjcolvar: Sorry, yes.  (I think I had changed this comment to be clearer but it got reverted. :slightly_frowning_face: )  
I was assuming the filename in this url would match `src/json/test_data/transcript-manifest-rendering.js` (except for the js/json difference).  Is that correct or am I linking things that shouldn't be?

Dananji: No these urls do not match within the /src/json directory.  I cleaned up that data and put these on maple. The files in the /src/json are for testing purposes. I'll change them, since it's good idea to have them renamed.

PR #99, opened 2021-10-19
Remove time tooltip over playhead

Dananji:
Before:

timetooltip-before

This PR uses `VideoJS` options when initializing the player instance to switch off the time tooltip over the playhead when hovering over the time rail. The fix in the player looks like this;

timetooltip


apertome (approved):
Looks good

PR #107, opened 2021-12-02
Parse 'start' prop in manifests, render accordingly in all components

Dananji:
Custom start in the manifest:
```
  start: {
    id: 'https://dlib.indiana.edu/iiif_av/lunchroom_manners/lunchroom_manners.json',
    type: 'SpecificResource',
    source: 'https://dlib.indiana.edu/iiif_av/lunchroom_manners/canvas/1',
    selector: {
      type: 'PointSelector',
      t: 180,
    },
  },
```
gets rendered as follows in the IIIFPlayer, StructuredNavigation, and Transcript components as follows;

custom-start


src/json/test_data/volleyball-for-boys.js
```diff
@@ -4,22 +4,29 @@ export default {
     'http://iiif.io/api/presentation/3/context.json',
   ],
   type: 'Manifest',
-  id:
-    'https://dlib.indiana.edu/iiif-av/iiif-player-samples/volleybal-for-boys/manifest',
+  id: 'https://dlib.indiana.edu/iiif-av/iiif-player-samples/volleybal-for-boys/manifest',
```
cjcolvar, line 7: It looks like this id is missing the second `l` in volleyball.  I think it would be nice to fix this throughout the file if possible.

src/services/iiif-parser.js
```diff
@@ -299,3 +289,42 @@ export function getPoster(manifest) {
...
+        currentCanvasIndex = getCanvasIndex(startProp.id);
+        return { type: 'C', canvas: currentCanvasIndex, time: 0 };
+      case 'SpecificResource':
+        currentCanvasIndex = getCanvasIndex(startProp.source);
+        let customStart = startProp.selector.t;
+        return { type: 'SR', canvas: currentCanvasIndex, time: customStart };
+    }
+  } else {
+    return null;
+  }
```
cjcolvar, line 329: This `else` should be unreachable because of the early return at the top of the method so can it be removed?

cjcolvar (approved):
This looks great!  I had two small suggestions which are minor and shouldn't block this from being merged.

src/services/iiif-parser.js
Dananji: Absolutely! :+1:

PR #118, opened 2022-01-12
Demo content

Dananji:


cjcolvar (commented):
Thanks for cleaning this up!  This is looking really good!

I wonder if there is a way to avoid the duplication of files between `public` and `docs`.  My concern with the duplication is avoiding issues where they get out of sync and needing to remember to update both of them.

Also I'm curious why some of the manifests have ids of `localhost` and others have `example.com`.

Dananji:
When `styleguidist` build the `docs` it's copying over the content in `public` directory to `docs`, so there's less chance of them being incosistent unless we make changes manually. Let me check whether this copying can be stopped, and it works properly without it.

The manifests in `public` is for demo purposes (with the `localhost` id), while the manifests in `json` -> `test_data` is for testing (with the `example.com` id). 
And the one manifest in `json` with the `localhost` id is to be used with the `manifest` prop in the `IIIFPlayer` component.

Dananji:
@cjcolvar I don't think it will work without the assets being copied over to the `docs` in the GitHub demo pages. So, do you think this is ready to be merged?

cjcolvar (approved):

PR #142, opened 2022-06-21
Bump dependencies

Dananji:
Fixes many of the security vulnerabilities.


cjcolvar:
Looks like the new dependencies require newer versions of node 14.15+

PR #143, opened 2022-06-27
Fix minor bugs in structure nav and file download

Dananji:
These minor bugs were identified while testing cookbook recipe manifests in the components.
This PR fixes the following bugs:
- Structure navigation was not working properly where there are `mediafragments` without an end time (assumes the duration of the media file is the end time)
- File download for alternative files in the manifest (under `rendering` property) were not working for cross-origin files


src/components/MediaPlayer/VideoJSPlayer.js
```diff
@@ -316,6 +318,7 @@ function VideoJSPlayer({
           type: 'switchItem',
         });
       } else {
+        console.log('handle ended with null');
```
cjcolvar, line 321: Did you intend to leave these `console.log` calls?

cjcolvar (approved):
Looks great!

src/components/MediaPlayer/VideoJSPlayer.js
Dananji: No I didn't! Good catch :100:

PR #153, opened 2023-01-27
Add queryparam support in demo, to work with IIIF cookbook recipes

Dananji:


cjcolvar (approved):
Exciting to see this mature and be able to be used with the cookbook!

PR #155, opened 2023-03-08
Caption support

Dananji:


cjcolvar (approved):
Looks good from what I can tell.

PR #170, opened 2023-04-26
Use placeholderCanvas to render poster image

### LINKED ISSUE #164: Render Poster Image in Media Player
# Description
Updating Ramp to read the appropriate reference to the poster image according to the Presentation 3 spec (placeholderCanvas instead of thumbnail), and render appropriately.

# Done Criteria
- [x] Ramp renders the image contain in the manifest and typed as a poster image appropriately within the media player

# Definition of Done
- Test in https://iiif-react-media-player.netlify.app/
- Displays the poster image referenced in `placeholderCanvas` property if available when the player reads a manifest. Examples from cookbook:
  - manifest with `placeholderCanvas`: https://iiif.io/api/cookbook/recipe/0013-placeholderCanvas/manifest.json => shows the image linked under `placeholderCanvas` property
  - manifest without: https://iiif.io/api/cookbook/recipe/0003-mvm-video/manifest.json => shows a black frame until playback starts

Dananji:
Closes #164


src/services/iiif-parser.js
```diff
@@ -324,13 +325,24 @@ export function getSegmentMap({ manifest }) {
...
+  let placeholderCanvas = parseManifest(manifest)
+    .getSequences()[0]
+    .getCanvasByIndex(canvasIndex)
+    .__jsonld['placeholderCanvas'];
+  if (placeholderCanvas) {
+    let annotations = placeholderCanvas['items'];
+    let items = parseAnnotations(annotations, 'painting');
+    if (items.length > 0) {
+      posterUrl = items[0].getBody()[0].id;
+    }
```
cjcolvar, line 341: If ramp assumes that the poster is a static image then it might be good to do a check here that the annotation body is type `Image` since placeholderCanvas can have any content type usually allowed in canvases.

Future work might be having ramp respect the width and height of placeholderCanvas since it can be different than the canvas itself.

cjcolvar (commented):
Looking good!  Just had one question.

cjcolvar (approved):
Looks good!

PR #188, opened 2023-05-24
Add previous/next section buttons to player icons

### LINKED ISSUE #176: Add Previous and Next Buttons to navigate between Canvases
# Description
When there are multiple canvases in a manifest, Ramp should allow for navigation between them via controls on the player's toolbar. With back and next buttons available when there is more than one file, users can navigate more quickly. VideoJS appears to have an icon for this under "previous-item".

# Definition of Done
- [ ] When more than one canvas/file is present in a manifest, "back" and "next" buttons appear next to the Play button
- [ ] Selecting the back button will load the preceding canvas media into the player
- [ ] Selecting the next button will load the following canvas media into the player
- [ ] Selecting Back on the first section will restart the section from the beginning
- [ ] Selecting Next on the last section will have no effect

# Mockup

Dananji:
Screenshot from 2023-05-24 14-09-00

Closes #176


cjcolvar (approved):
This looks great!  The only question I had was if the previous/next buttons should only appear when there are previous or next canvases to change to?  I noticed that when on the first canvas the previous button sets the player to the beginning of the section while the next button on the last canvas does nothing.  Maybe these buttons could be rendered but disabled when they aren't actionable?  But if how you coded things is the desired behavior then feel free to go ahead and merge.

cjcolvar (commented):
This PR changes locations of some files. #177 modifies some of the same files, does it need to be merged first?

Dananji:
> The only question I had was if the previous/next buttons should only appear when there are previous or next canvases to change to?

Yes, previous/Next buttons are displayed only when there are canvases to switch

>  I noticed that when on the first canvas the previous button sets the player to the beginning of the section while the next button on the last canvas does nothing. Maybe these buttons could be rendered but disabled when they aren't actionable?

When on first canvas the previous button restarts playback from start and it is conveyed in the title of the button when hovered so that the user is not confused. This was unusual to me at first because I haven't paid attention to this before, but this seems to be the common behavior of major players out there :)
As for the last canvas Jon's expectation was that the next button does nothing, but I like the idea of disabling it since it doesn't do anything in that use-case. Maybe we could revisit this in the future :+1:

PR #192, opened 2023-05-30
Metadata component

### LINKED ISSUE #172: Create metadata component for Ramp
# Description

In order to display media object metadata from the IIIF manifest in Ramp, create a media object component that can be enabled on the page.

Example record with lots of metadata: https://avalon-dev.dlib.indiana.edu/media_objects/6t053f984/

General look and feel should parallel what is in Avalon: Labels in bold, metadata indented underneath; multiple entries concatenated together with semi-colon.

# Definition of Done
- [ ] All metadata fields in the IIIF manifest should display except title
- [ ] Metadata should be displayed in the same order it occurs in the IIIF manifest. 
- [ ] HTML tags in the metadata should be rendered as HTML.
- [ ] Write wiki docs for the new component

# Questions
- Do multiple values need to be concatenated?
No, values will come through as an array of strings that can be rendered as a list

- For the component, can title be included but not displayed in Avalon? (for more general-purpose use)
Already in label value in manifest; could be included in metadata but suppressed with a prop in the code

Dananji:
Metadata component displays the metadata from a manifest if available. There are 2 props;
- displayTitle: defaults to `true` to display title within the metadata display
- showHeading: defaults to `true` to display header element in the display

In Netlify demo site with Transcripts and Details in a tabbed panel;
Screenshot from 2023-05-31 14-18-59

When metadata is not available in the manifest:
Screenshot from 2023-05-31 14-17-41

As a raw component in the library (from Styleguidist docs):
Screenshot from 2023-05-30 16-36-42

Closes #172


src/services/iiif-parser.js
```diff
@@ -440,3 +440,28 @@ export function getRenderingFiles(manifest, canvasIndex) {
...
 }
+
+
+/**
+ * @param {Object} manifest
+ * @return {Array} list of key value pairs for each metadata item in the manifest
+ */
+export function parseMetadata(manifest) {
+  try {
+    const metadata = parseManifest(manifest).getMetadata();
```
cjcolvar, line 451: Do you know if `manifesto` strips unallowed HTML tags from the metadata values?  From what I can tell it isn't.
https://iiif.io/api/presentation/3.0/#45-html-markup-in-property-values

cjcolvar (commented):
This looks really good.  The only question/concern I have is around HTML markup in metadata values.

src/services/iiif-parser.js
Dananji: Yes, it seems like it doesn't strip the HTML. Maybe it's good to have this inside the parsing code. I'll check that out :+1:

Dananji: Added HTML sanitation to the metadata using `sanitize-html` JS library

cjcolvar (approved):
Looks good.

PR #193, opened 2023-05-31
Fix bug when switching canvas using previous/next buttons

Dananji:
Bug fix for switching between canvases using previous/next buttons to update start time of the player

### DIFF
--- a/src/components/MediaPlayer/VideoJSPlayer.js
+++ b/src/components/MediaPlayer/VideoJSPlayer.js
@@ -222,7 +222,7 @@ function VideoJSPlayer({
    */
   React.useEffect(() => {
     if (isClicked && canvasIndex !== cIndex) {
-      switchPlayer(canvasIndex);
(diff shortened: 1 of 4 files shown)


src/components/MediaPlayer/VideoJSPlayer.js
```diff
@@ -222,7 +222,7 @@ function VideoJSPlayer({
    */
   React.useEffect(() => {
     if (isClicked && canvasIndex !== cIndex) {
-      switchPlayer(canvasIndex);
+      switchPlayer(canvasIndex, false);
```
cjcolvar, line 225: Why is `fromStart` false in this case?

Dananji: Because the current time of the player is set to the start value of the clicked structure item. By setting this flag to true, it will always set the time to the start of the new canvas within the function, which is the case for all the other cases except for when using structure navigation.

PR #199, opened 2023-06-14
UI updates to Ramp media player and videojs

Dananji:


cjcolvar (approved):
Looks great!  The refactor introducing `isVideo` is a real nice cleanup!

PR #201, opened 2023-06-23
Restructure

### LINKED ISSUE #197: Code Restructuring
# Description
Cleanup tasks in Ramp code.

# Done Looks Like
- [ ] Go through testing data and remove duplicate files (both Styleguidist and Netlify)
- [x] Clean out unused components from the starter template
- [ ] Utility components (List component) are out of standard app structure and should be re-arranged/re-organized with the repository structure

Dananji:
Closes #197


src/components/MediaPlayer/MediaPlayer.md
```diff
@@ -5,19 +5,13 @@ MediaPlayer component provides a player that facilitates both audio and video me
...
-
-```jsx padded
-import IIIFPlayer from '../IIIFPlayer/IIIFPlayer';
-import config from '../../../env.js';
+import { IIIFPlayer, MediaPlayer } from '@samvera/ramp';
 
-<IIIFPlayer
-  manifestUrl={`${config.url}/manifests/${config.env}/lunchroom_manners.json`}
->
+<IIIFPlayer manifestUrl={manifest_url} manifest={manifest}>
```
cjcolvar, line 14: What should `manifestUrl` and `manifest` be?  Should you only use one at a time or both?
I think it would be good to standardize the documented usage and/or clarify what each prop is and why you'd use one or the other or both.
FWIW This looks the same as `StructuredNavigation.md`, but `IIIFPlayer.md` and `MetadataDisplay.md` only use `manifest` and explicitly set it to `lunchroom_manners.js`.

cjcolvar (commented):
Looks good.  Nice cleanup!  I had one question about props for `IIIFPlayer` used in documentation.

src/components/MediaPlayer/MediaPlayer.md
Dananji: In the `IIIFPlayer.md` the props are explained. It mentions that the `manifest` prop takes precedence over the `manifestUrl`.

And in the `MediaPlayer.md` and `StructuredNavigation.md` I removed the actual code because it replicate the components displayed at top with `IIIFPlayer` component. Let me change it so it's not confusing and looks the same as the others. Thanks for spotting this :100:

cjcolvar (approved):
I missed the definition of those props in IIIFPlayer.md.  Referring back to it in the other docs is an improvement.  Thanks!

PR #206, opened 2023-07-12
Supplemental files

cjcolvar:
Add component to show supplemental files (renderings) from manifest and canvas level.

TODO: Write tests, fix CSS (padding/margin in tab), refactor canvas label fetching code?


Dananji:
This looks great :100: I had only one comment on the SupplementalFiles component's conditional rendering part.

src/components/SupplementalFiles/SupplementalFiles.js
```diff
@@ -0,0 +1,77 @@
...
+  return (
+    <div data-testid="supplemental-files" className="ramp--supplemental-files">
+      {showHeading && (
+        <div className="ramp--supplemental-files-heading" data-testid="supplemental-files-heading">
+          <h4>Files</h4>
+        </div>
+      )}
+      <div className="ramp--supplemental-files-display-content" data-testid="supplemental-files-display-content">
+        {Array.isArray(manifestSupplementalFiles) && manifestSupplementalFiles.length > 0 && ( <h4>{itemHeading}</h4> )}
+        {Array.isArray(manifestSupplementalFiles) && manifestSupplementalFiles.length > 0 && (
```
Dananji, line 43: Here, instead of doing the check twice is it possible to do it like this?
```
{Array.isArray(manifestSupplementalFiles) && manifestSupplementalFiles.length > 0 && (
  <React.Fragment>
    <h4>{itemHeading}</h4>
    <dl key={'item-files'}>
      { manifestSupplementalFiles.map((file, index) => {
         return (
           <dd key={index}>
             <a href={file.id} key={index} onClick={e => handleDownload(e, file)}>{file.label}</a>
           </dd>);
        })
     }
    </dl>
  </React.Fragment>
)}
```

cjcolvar: Thanks for the suggestion!  It worked. :tada:

Dananji (commented):
This looks great :tada:

Dananji:
I added new tests and rebased the branch from `main` after merging machine-generated transcript work. Changed some of the code which broke from the rebase. And wrote some documentation for the styleguidist docs.

src/components/SupplementalFiles/SupplementalFiles.js
```diff
@@ -0,0 +1,97 @@
...
+
+      let annotations = getSupplementingFiles(manifest);
+      let canvasFiles = renderings.canvas;
+      canvasFiles.map((canvas, index) => canvas.files = canvas.files.concat(annotations[index].files));
+      canvasFiles = canvasFiles.filter(canvasFiles => canvasFiles.files.length > 0);
+      setCanvasSupplementalFiles(canvasFiles);
+
+      if (canvasFiles?.length > 0 && manifestFiles?.length > 0) {
+        setHasFiles(true);
+      }
```
cjcolvar, line 29: What do you think about converting this into a function instead of storing it in the state?  Would that even work?

Dananji: I guess it will work even without storing it in the state because the variables that determine its values are state variables which get updated when props change. Let me check.

Dananji: Yes, it works like that. I changed the code.

Dananji (approved):
:100:

PR #211, opened 2023-08-07
Fix VideoJS file download bug

Dananji:
Fix for file download icon with the changed parsing implemented in the SupplementalFiles component.


cjcolvar (approved):

PR #216, opened 2023-08-15
Transcripts fix

Dananji:
In this PR;
- Fix delay in transcript loading when switching between canvases
- Cache transcript data in Transcript component state to avoid delays when switching between transcripts and canvases (content in the transcript file is requested for initial render for each transcript, and then cached in the state for any re-renders. State is cleared when transcript is unmounted)


src/services/transcript-parser.js
```diff
@@ -18,114 +18,132 @@ const TRANSCRIPT_MIME_TYPES = [
...
-    .catch(function () {
-      return [{ ...item, validity: TRANSCRIPT_VALIDITY.invalidURL }];
+    .catch(function (err) {
+      console.error('Error fetching manifest, ', manifestURL);
+      return [];
     });
   return data;
 }
 
+export async function sanitizeTranscripts(transcripts) {
```
cjcolvar, line 94: I'm a bit confused as to what all this method is doing.  Could you please add a JSDoc method description comment and/or break it down into smaller methods?

cjcolvar (approved):
This looks good as far as I can tell.  I had to learn some React magic to understand what appeared to be duplicate `useEffect` declarations but I think I understand it now.  There was one method that I wasn't able to fully grasp.  It would be nice if that was made clearer but technically there isn't any reason to block this PR so I'm going ahead and approving it.

src/services/transcript-parser.js
Dananji: I was not able to break it down to smaller methods without breaking the code :disappointed:. 

So I added comments to it as much as I could to explain what each code block does.

cjcolvar (approved):

PR #218, opened 2023-08-18
Add top level auto-advance toggle component.

cjcolvar:
image

### DIFF
--- a/src/components/AutoAdvanceToggle/AutoAdvanceToggle.js
+++ b/src/components/AutoAdvanceToggle/AutoAdvanceToggle.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import { useManifestState, useManifestDispatch } from '../../context/manifest-context';
 import './AutoAdvanceToggle.scss';
 
--- a/src/components/AutoAdvanceToggle/AutoAdvanceToggle.js
+++ b/src/components/AutoAdvanceToggle/AutoAdvanceToggle.js
@@ -20,9 +20,18 @@ const AutoAdvanceToggle = ({ label = "Autoplay", showLabel = true }) => {
   return (
     <div data-testid="auto-advance" className="ramp--auto-advance">
       {showLabel && (
-        <span className="ramp--auto-advance-label" data-testid="auto-advance-label" for="auto-advance-toggle">{label}</span>
+        <span
+          className="ramp--auto-advance-label"
+          data-testid="auto-advance-label"
+          htmlFor="auto-advance-toggle"
+          id="auto-advance-toggle-label"
+        >
+          {label}
+        </span>
       )}
-      <label className="ramp--auto-advance-toggle">
+      <label className="ramp--auto-advance-toggle"
+        aria-labelledby='auto-advance-toggle-label'
+      >
         <input
           data-testid="auto-advance-toggle"
           name="auto-advance-toggle"
--- a/src/components/AutoAdvanceToggle/AutoAdvanceToggle.js
+++ b/src/components/AutoAdvanceToggle/AutoAdvanceToggle.js
@@ -6,13 +6,12 @@ import './AutoAdvanceToggle.scss';
 const AutoAdvanceToggle = ({ label = "Autoplay", showLabel = true }) => {
   const { manifest } = useManifestState();
 
-  const [autoAdvance, setAutoAdvance] = React.useState();
+  const [autoAdvance, setAutoAdvance] = React.useState('');
 
   React.useEffect(() => {
     if (manifest) {
       //Parse manifest to see if auto_advance behavior present at manifest level then save into state
       const manifestParsed = parseManifest(manifest);
-      console.log(manifestParsed.getBehavior());
       const autoAdvanceBehavior = manifestParsed.getBehavior() === "auto-advance";
       setAutoAdvance(autoAdvanceBehavior);
     }
@@ -23,7 +22,7 @@ const AutoAdvanceToggle = ({ label = "Autoplay", showLabel = true }) => {
       {showLabel && (
         <span className="ramp--auto-advance-label" data-testid="auto-advance-label">{label}</span>
       )}
-      <input type="checkbox" data-toggle="toggle" data-size="mini" name="ramp--auto-advance-toggle" checked={autoAdvance} onChange={e => setAutoAdvance(e.target.checked)} />
+      <input type="checkbox" data-testid="auto-advance-toggle" data-toggle="toggle" data-size="mini" name="ramp--auto-advance-toggle" checked={autoAdvance} onChange={e => setAutoAdvance(e.target.checked)} />
     </div>
   );
 };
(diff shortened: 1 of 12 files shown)


src/components/AutoAdvanceToggle/AutoAdvanceToggle.test.js
```diff
@@ -0,0 +1,54 @@
...
+    });
+    test('auto advance is toggled on click', () => {
+      const toggle = screen.queryByTestId('auto-advance-toggle');
+      fireEvent.click(toggle);
+      expect(toggle.checked).toEqual(true);
+    });
+  });
+  describe('with manifest with auto-advance behavior', () => {
+    beforeEach(() => {
+      const AutoAdvanceWithManifest = withManifestProvider(AutoAdvanceToggle, {
```
masaball, line 34: This should probably be renamed to `AutoManifest` or `AutoAdvanceManifest`. Unsure what my original logic with the name was, but this naming is inaccurate and potentially confusing.

masaball (commented):
Looks good! Have a couple suggestions in the test suite, but nothing else jumped out.

src/components/AutoAdvanceToggle/AutoAdvanceToggle.test.js
```diff
@@ -0,0 +1,54 @@
...
+      });
+      render(<NonAutoManifest />);
+    });
+    test('renders auto advance toggle successfully', () => {
+      expect(screen.queryByTestId('auto-advance')).toBeInTheDocument();
+      expect(screen.queryByTestId('auto-advance-label')).toBeInTheDocument();
+      expect(screen.queryByTestId('auto-advance-toggle')).toBeInTheDocument();
+    });
+    test('auto advance is disabled by defulat', () => {
+      const toggle = screen.queryByTestId('auto-advance-toggle');
```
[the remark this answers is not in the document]
cjcolvar: I tried DRYing this up but couldn't find any pattern that works.

cjcolvar:
The last change I made wires the toggle up to the manifest state and the player will check that when a canvas ends playing before moving on to the next canvas.  This is a fairly significant change in behavior for the player as it autoplayed by default before.  We could add `auto-advance` behavior to all of avalon's manifests but I'm not sure how hard that would be for other adopters.  Maybe we could add a prop to IIIFPlayer to opt-in to this new behavior and deprecate the old behavior to be removed in a future major release?

cjcolvar:
There is a `no-auto-advance` behavior in the IIIF Presentation apec and it is supposed to be default.  So I think we could consider Ramp's current behavior a bug.  I'll go ahead and make the change for `auto-advance` to default to `false` unless it is present in the manifest.  With this change it would probably be good to cut a major release and call out the change in behavior so adopters can modify their manifests accordingly.

src/components/AutoAdvanceToggle/AutoAdvanceToggle.test.js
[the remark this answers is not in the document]
Dananji: This could be done by initializing `toggle` before the `beforeEach` block and then assigning the value inside the `beforeEach` block.
Example: https://github.com/avalonmediasystem/react-structural-metadata-editor/blob/7fec5982340fda3a49058b85ba47c40363d18371/src/components/__tests__/ListItemControls.test.js#L124

Dananji:
Looks good overall, I had a couple of questions and suggestions.

src/components/AutoAdvanceToggle/AutoAdvanceToggle.js
```diff
@@ -0,0 +1,37 @@
...
+          type="checkbox"
+          checked={autoAdvance}
+          onChange={e => manifestDispatch({ autoAdvance: e.target.checked, type: "setAutoAdvance" })}
+        />
+        <span className="slider round"></span>
+      </label>
+    </div>
+  );
+};
+
```
Dananji, line 36: Could we add typechecking for the component props?

src/components/IIIFPlayerWrapper.js
```diff
@@ -28,6 +29,17 @@ export default function IIIFPlayerWrapper({
     }
   }, []);
 
+  React.useEffect(() => {
+    if (manifest) {
+      //Parse manifest to see if auto-advance behavior present at manifest level then save into state
+      const manifestParsed = parseManifest(manifest);
+      const autoAdvanceBehavior = manifestParsed.getProperty("behavior")?.includes("auto-advance");
```
Dananji, line 36: I think it would be good to have all the `manifesto.js` parsing code in the `iiif-parser` module.
Maybe we could move this into a new function in the parser? What do you think?

src/components/AutoAdvanceToggle/AutoAdvanceToggle.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import { useManifestState, useManifestDispatch } from '../../context/manifest-context';
+import './AutoAdvanceToggle.scss';
+
+const AutoAdvanceToggle = ({ label = "Autoplay", showLabel = true }) => {
+  const { manifest, autoAdvance } = useManifestState();
```
Dananji, line 6: Here, `manifest` can be removed since we don't use it in this component anymore?

src/components/AutoAdvanceToggle/AutoAdvanceToggle.test.js
```diff
@@ -0,0 +1,53 @@
...
+import AutoAdvanceToggle from "./AutoAdvanceToggle";
+import manifest from '@TestData/lunchroom-manners';
+import autoAdvanceManifest from '@TestData/multiple-canvas-auto-advance';
+import { withManifestProvider } from '../../services/testing-helpers';
+
+describe('AutoAdvanceToggle', () => {
+  describe('with manifest without auto-advance behavior', () => {
+    beforeEach(() => {
+      const NonAutoManifest = withManifestProvider(AutoAdvanceToggle, {
+        initialState: { manifest, canvasIndex: 0 },
```
Dananji, line 12: When I ran the test suite locally I got the following in the console;
```
    console.error
      Warning: A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://reactjs.org/link/controlled-components
          at input
          at label
          at div
          at AutoAdvanceToggle
```
This is triggered by not initializing `autoAdvance` in the `initialState` object passed into the state when setting up the component for testing. Setting `autoAdvance: false` in `initialState` gets rid of the warning.
This is not causing any tests to fail and I wanted to make a note of it.

src/components/AutoAdvanceToggle/AutoAdvanceToggle.js
cjcolvar: Good catch!

Dananji:
This looks great :100: 
I noticed just one thing below. After fixing it we can merge this :+1:

src/components/IIIFPlayerWrapper.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import { useManifestDispatch } from '../context/manifest-context';
 import PropTypes from 'prop-types';
-import { parseManifest } from 'manifesto.js';
+import { parseAutoAdvance } from 'manifesto.js';
```
Dananji, line 4: I think this import should be from `iiif-parser` module?

Dananji (approved):

src/components/IIIFPlayerWrapper.js
cjcolvar: Surprised the tests passed with that error!

PR #219, opened 2023-08-28
Add handling for empty canvases within playlist manifests

### LINKED ISSUE #213: Display Empty Canvases as Inaccessible Items
# Description
Items that are inaccessible to users should still be listed as items within a playlist.

# Done Looks Like
- Items in a playlist manifest without content in the canvas are shown in the structure display as "Inaccessible Item" in the structured navigation
- placeHolderCanvas should be displayed from the manifest, but Ramp player controls should not be accessible to user (since the video can't be played, user cannot create playlist markers, etc.)

masaball:
Closes #213 

image

### DIFF
--- a/src/services/iiif-parser.js
+++ b/src/services/iiif-parser.js
@@ -20,7 +20,15 @@ export function canvasesInManifest(manifest) {
     .getSequences()[0]
     .getCanvases()
     .map((canvas) => {
-      return canvas.id;
+      try {
+        let sources = canvas
+          .getContent()[0]
+          .getBody()
+          .map((source) => source.id);
+        return { canvasId: canvas.id, isEmpty: false };
+      } catch (error) {
+        return { canvasId: canvas.id, isEmpty: true };
+      }
     });
   return canvases;
 }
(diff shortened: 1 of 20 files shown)


src/services/iiif-parser.js
```diff
@@ -373,12 +368,10 @@ export function getCustomStart(manifest) {
...
-      .map(function (c) {
-        return c;
-      })
-      .indexOf(canvasId);
+    const currentCanvasIndex = canvasesInManifest(manifest)
+      .findIndex((c) => {
+        return c.canvasId === canvasId;
+      });
     return currentCanvasIndex;
   };
```
cjcolvar, line 376: This pattern seems to be coming up a few times.  Maybe it could be extracted into a helper function in a future PR?

cjcolvar (approved):
Looks great!

src/services/iiif-parser.js
Dananji: Yes :100: That sounds good. I'll make a note.

PR #223, opened 2023-08-30
Refactor manifest setting and fetching to ManifestProvider instead of IIIFPlayerWrapper

cjcolvar:
Not sure if this is a good direction or not.


cjcolvar:
Duplicate of #222

PR #231, opened 2023-09-08
Fix broken scrubbing due to time tooltip styling

Dananji:
Before:
image
With the time tooltip on top of the time rail, scrubbing doesn't work because the progress bar doesn't register the mouse click event.

After:
image


cjcolvar (approved):
Nice catch!

Issue #184, opened 2023-05-24
i18n support


joncameron:
React has a native localization support; this will require review of the code but should be straightforward. For the custom components, we can use localization and do proper labeling. Video.js elements may be less straightforward, but it should be possible to override these within the Ramp context as well. We can investigate more about how to do this in React.

There is also currently no language selection or handling in Ramp if multiple languages are available in a manifest; this can be investigated as well.

# Done Looks Like
- [x] Determine code changes needed to be made to add i18n support
- [x] Investigate possible handling of multi-language values in manifests
- [x] Investigate current methods to do this in React/Video.js
- [x] Create new issues as needed

Dananji:
Correction: React doesn't have native localization support.

To support internationalization in Ramp, we need to implement the following;
1. Reading multi language value in a Manifest: 
Cookbook recipe: https://iiif.io/api/cookbook/recipe/0006-text-language/
In the current implementation Ramp reads the first value in a language map. To support multiple languages we need changes to the parser to read multiple values and display them accordingly. 
To change language throughout the components we can implement a language prop for `IIIFPlayer` component, which wraps all the other components. Use it when parsing the Manifest and default to English or first available value when the given language is not in the language map.
Language selection in other viewers: Both UV and Mirador has a language selection option where it gives a specific set of languages and a user can select a language from it, these language choices are not affected by the languages in the Manifest.

2. Add translations to text values in the UI using [`react-i18next` library](https://www.npmjs.com/package/react-i18next). Mirador uses the same library with the ability to add new languages using configs. This implementation impacts all the components with text hard-coded in them and the error messages the parser sets. As a first step we could implement support for a couple of languages, and then look into adding support for adding new languages in the future.

3. In VideoJS: use `language` option to set language in the player. From a quick look at the code, to implement this specify language in the VideoJS config and then import the lang file from VideoJS library, set the language on the videojs instance.
Example: https://codesandbox.io/s/silly-butterfly-80pibt?file=/src/video.js:96-140

```
videoJsOptions: { ...options, language: "de" }

In VideoJSPlayer module:
  import de from 'video.js/dist/lang/de.json';
  import fr from 'video.js/dist/lang/fr.json';
  import en from 'video.js/dist/lang/en.json';
  ...
  
  const LANG_MAP = { "de": de, "fr": fr, "en": en };
  ...

    let languageJSON = LANG_MAP[options.language];
    let newPlayer;
    if (playerRef.current != null) {
      languageJSON
[...]

PR #235, opened 2023-09-11
Add i18n support with language optin in VideoJS

Dananji:
English:

image

French:

image

Part of #184


src/components/MediaPlayer/MediaPlayer.js
```diff
@@ -210,6 +210,7 @@ const MediaPlayer = ({ enableFileDownload = false, enablePIP = false }) => {
     poster: isVideo ? getPoster(manifest, canvasIndex) : null,
     controls: true,
     fluid: true,
+    language: "fr", // TODO:: fill this information from props
```
cjcolvar, line 213: Should this be switched to `en` or removed?

src/components/MediaPlayer/VideoJS/VideoJSUtils.js
```diff
@@ -0,0 +1,15 @@
...
+import VideoJSPreviousButton from './components/js/VideoJSPreviousButton';
+// import vjsYo from './vjsYo';
+
+/** VideoJS language files */
+import de from "video.js/dist/lang/de.json";
+import en from "video.js/dist/lang/en.json";
+import fr from "video.js/dist/lang/fr.json";
+import it from "video.js/dist/lang/it.json";
+
+export const LANG_MAP = { "de": de, "fr": fr, "it": it, "en": en };
```
cjcolvar, line 15: If the language used in the interface is based upon a prop then it would be nice to dynamically load the language json file if that's possible.  If not, I wonder if it might be helpful to load all of the languages here if that doesn't have too large of a performance impact.  If they are all loaded then all of the options are available to the implementor without having to modify ramp.

src/components/MediaPlayer/MediaPlayer.js
Dananji: Good catch! It should be `en` for now.

cjcolvar (approved):
Neat!  I wonder if we'll want to cache any of this importing (or if it already magically happens) so it doesn't have to load with each video.js instantiation.  We can wait and see if this actually is a performance issue though.

PR #242, opened 2023-09-13
Player respects auto advance toggle

### LINKED ISSUE #214: Add Auto-advance Component
# Description
From a code perspective it makes sense to have the auto-advance toggle to be a part of Ramp, since the state machine for this is on the Ramp side. It's easier to hook this up to the player state within Ramp rather than via Avalon.

# Done Looks Like
- [x] An auto-advance toggle component for Ramp is added
- [x] Can be a part of structured navigation in the UI
- [x] Auto-advance behavior depends on the state of the toggle

cjcolvar:
Resolves #214


Dananji (approved):
Looks good :+1:

PR #243, opened 2023-09-15
Add check for empty item to getItemId

masaball:
[The behavior encountered](https://github.com/avalonmediasystem/avalon/issues/5350) by Avalon with Ramp components not loading was because of top level ranges having an empty item section: `item: [ ]`. The parser would attempt to grab the item's ID and the parsing would fail, causing all of the components to stop processing.

This PR adds a condition to check the length of the item array before grabbing the ID value.


Dananji (approved):
Looks good :+1:

PR #245, opened 2023-09-20
Add marker component

cjcolvar:
Includes rebase of annotation-service-markers branch.


Dananji:
With form not collapsed;
image

Create marker form, time is auto filled from the current playhead time in the player when `Add new marker` button is clicked;
image

Timestamp validation;
image

Feedback upon failed submission;
image

PR #248, opened 2023-09-22
Fix player css for zooming in demo site

Dananji:
The issue with icons in the control bar we were seeing when zooming-in was happening only in the demo site, because it sets the player width to 60%. The issue was not with VideoJS CSS. Example screenshots below where the player's width is not adjusted show the full control bar at 240% zoom level in the browser. 

GitHub documentation site:
image

avalon-dev:
image


cjcolvar (approved):
Good catch!

PR #250, opened 2023-09-28
Replace videojs-hotkeys with builtin hotkeys function

masaball:
Videojs-hotkeys was not interfacing correctly between Avalon and embedded Ramp. This could be an issue encountered by other implementers. Rather than cobble together a fix on the Avalon side and documenting that for others, it felt better to try to change the approach in Ramp so that hopefully workarounds would not be necessary. 

To that end, recent versions of VideoJS have added a builtin hotkey ability so we are going to try switching to that. This currently requires building the hotkeys out from scratch because the defaults do not include all of the functionality we require and adding any customization disables the defaults. It may be worthwhile exploring contributing "seek" and "volume" hotkey components back upstream to be added to the VideoJS default behaviors but that is a future discussion.

### DIFF
--- a/src/components/MediaPlayer/MediaPlayer.js
+++ b/src/components/MediaPlayer/MediaPlayer.js
@@ -244,6 +244,70 @@ const MediaPlayer = ({ enableFileDownload = false, enablePIP = false }) => {
       ? playerConfig.sources[srcIndex]
       : playerConfig.sources,
     tracks: playerConfig.tracks,
+    userActions: {
+      hotkeys: function(event) {
+        // event.which key code values found at: https://css-tricks.com/snippets/javascript/javascript-keycodes/
+
+        // Prevent default browser actions so that page does not react when hotkeys are used.
+        // e.g. pressing space will pause/play without scrolling the page down.
+        event.preventDefault();
+
+        // Space and k toggle play/pause
+        if (event.which === 32 || event.which === 75) {
+          if (this.paused()) {
+            this.play();
+          } else {
+            this.pause();
+          }
+        }
+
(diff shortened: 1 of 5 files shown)


Dananji:
This looks great :100:

src/components/MediaPlayer/MediaPlayer.js
```diff
@@ -244,6 +244,70 @@ const MediaPlayer = ({ enableFileDownload = false, enablePIP = false }) => {
...
+        }
+
+        // Right arrow seeks 5 seconds ahead
+        if (event.which === 39) {
+          this.currentTime(this.currentTime() + 5);
+        }
+
+        // Left arrow seeks 5 seconds back
+        if (event.which === 37) {
+          this.currentTime(this.currentTime() - 5);
```
Dananji, line 297: Do we need a check here for the duration of the media in case the `currentTime` after the 5 second adjustment is either greater than the duration or less than 0 or the alternate start time of the media?
Or is it getting handled on the VideoJS end of things?

Dananji (approved):

src/components/MediaPlayer/MediaPlayer.js
masaball: It looks like it is getting handled on the VideoJS end of things. Trying to seek back past the start time of a video/segment gets set to the start time. Seeking past the end of a video/segment either triggers the autoplay of the next section or, if there are no more sections, triggers the "Transcripts undefined" error we've recently been seeing which is unrelated to this specific seeking behavior. 

Neither case triggers any other error messages.

PR #251, opened 2023-09-28
Make the volume slider horizontal and sticky for audio player

Dananji:
Video player:
image

Audio player:
image


cjcolvar (approved):
Looks great!

PR #257, opened 2023-10-05
Prevent retrieving currentTime when player is disposed

masaball:
This PR also includes a commit that removes videojs-hotkeys initialization code that was accidentally reintroduced after we stopped using that plugin.


cjcolvar (approved):
Good catch on these!

PR #268, opened 2023-10-16
Make canvas-level items w/o mediafragment into spans instead of buttons

Dananji:
Canvas w/o mediafragment information displayed as a span:
image

Canvas with medifragment information displayed as a button;
image

### DIFF
--- a/src/components/StructuredNavigation/NavUtils/List.test.js
+++ b/src/components/StructuredNavigation/NavUtils/List.test.js
@@ -1,12 +1,11 @@
 import React from 'react';
 import List from './List';
 import { render, screen } from '@testing-library/react';
-import manifest from '@TestData/lunchroom-manners';
 import { withManifestAndPlayerProvider } from '../../../services/testing-helpers';
 
 describe('List component', () => {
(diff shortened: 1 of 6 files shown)


src/components/StructuredNavigation/NavUtils/List.test.js
```diff
@@ -144,19 +69,60 @@ describe('List component', () => {
       });
       render(<ListWithManifest />);
     });
-    test('renders successfully', () => {
+    test('sstructures uccessfully', () => {
```
cjcolvar, line 72:
```suggestion
    test('renders successfully', () => {
```

cjcolvar (commented):
This looks good but I wonder if it makes sense to rename `SectionButton` to something else since it can now be a button or a span.  Maybe `SectionHeading`?

Dananji:
> This looks good but I wonder if it makes sense to rename `SectionButton` to something else since it can now be a button or a span. Maybe `SectionHeading`?

Yes, sounds good! I'll make that change.

cjcolvar (approved):

PR #275, opened 2023-10-30
Add active section

Dananji:
Mark active section when there are;

Canvas with child timespans:
image

Canvas without child timespans:
image

Canvas without children (timespans/divs):
image


cjcolvar (commented):
This looks good as far as I can tell.  I noticed that there are a number of blocks that have been commented out.  Can those be removed?

cjcolvar (approved):
Thanks for cleaning that up!

PR #277, opened 2023-11-01
Display placeholder canvas message for all empty manifests

Dananji:
image

### DIFF
--- a/src/components/MediaPlayer/VideoJS/components/js/VideoJSCurrentTime.js
+++ b/src/components/MediaPlayer/VideoJS/components/js/VideoJSCurrentTime.js
@@ -49,7 +49,7 @@ function CurrentTimeDisplay({ player, options }) {
 
   let initTimeRef = React.useRef(options.currentTime);
   const setInitTime = (t) => {
-    initTimeRef.current = 0;
+    initTimeRef.current = t;
   };
 
   player.on('timeupdate', () => {
--- a/src/components/MediaPlayer/VideoJS/components/js/VideoJSCurrentTime.js
+++ b/src/components/MediaPlayer/VideoJS/components/js/VideoJSCurrentTime.js
@@ -47,12 +47,23 @@ function CurrentTimeDisplay({ player, options }) {
 
   const [currTime, setCurrTime] = React.useState(player.currentTime());
 
+  let initTimeRef = React.useRef(options.currentTime);
+  const setInitTime = (t) => {
+    initTimeRef.current = 0;
+  };
+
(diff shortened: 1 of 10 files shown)


src/components/MediaPlayer/VideoJS/components/js/VideoJSProgress.js
```diff
@@ -233,7 +236,7 @@ function ProgressBar({ player, handleTimeUpdate, times, options }) {
 
   player.on('timeupdate', () => {
     if (player.isDisposed()) return;
-    const curTime = player.currentTime();
+    const curTime = initCurrentTime || player.currentTime();
```
cjcolvar, line 239: If `initCurrentTime` is passed as a prop does it get updated somehow or does it stay static?  I'm wondering if this change blocks the progress bar from updating since it looks like `player.currentTime()` would never get returned if `initCurrentTime` is set.

src/components/MediaPlayer/VideoJS/components/js/VideoJSCurrentTime.js
```diff
@@ -47,12 +47,23 @@ function CurrentTimeDisplay({ player, options }) {
 
   const [currTime, setCurrTime] = React.useState(player.currentTime());
 
+  let initTimeRef = React.useRef(options.currentTime);
+  const setInitTime = (t) => {
+    initTimeRef.current = 0;
```
cjcolvar, line 52: Should this set to `t` instead of `0`?
```suggestion
    initTimeRef.current = t;
```

src/components/MediaPlayer/VideoJS/components/js/VideoJSProgress.js
```diff
@@ -184,6 +186,10 @@ function ProgressBar({ player, handleTimeUpdate, times, options }) {
 
   const isMultiSourced = options.targets.length > 1 ? true : false;
 
+  let initTimeRef = React.useRef(initCurrentTime);
+  const setInitTime = (t) => {
+    initTimeRef.current = 0;
```
cjcolvar, line 191: Should this set to t instead of 0?
```suggestion
    initTimeRef.current = t;
```

cjcolvar (commented):
I'm a little confused about what problem `initTimeRef` is solving and how it is solving it but if it is working than I don't necessarily need to fully understand it.  It also seems like the `timeupdate` handling in `VideoJSCurrentTime` and `VideoJSProgress` could be extracted into a common method but that doesn't need to be done within the scope of this PR and could be done later if they end up really being the same.

PR #279, opened 2023-11-07
Fix missing canvas console error

masaball:
0 evaluates as falsy in javascript, so we need to explictly check for it in our if/else check when grabbing canvasindex.


Dananji (approved):
Nice catch! Looks good :+1:

PR #282, opened 2023-11-08
Add scrollbar override to enable constant display

masaball:
This change is made in the main.scss file because the override defines a custom scrollbar design. If we were to limit this CSS to only one component (StructuredNavigation) the rest of the components would use the browser default scroll bar which may not match the styling defined here. For visual consistency within Ramp, we need to apply the scrollbar rule to all scrollbars in the application.


Dananji (approved):
Looks good!

PR #284, opened 2023-11-09
Fix structure display for section items with Canvas references

Dananji:
When playhead is at a timestamp not included in a timespan in the structure:
image

When playhead is at a timestamp included in a timespan in the structure:
image


masaball (approved):
Looks good!

PR #285, opened 2023-11-13
Keep focus on next/previous buttons when using keyboard navigation

Dananji:
Use a state variable in the state management called `playerFocusElement` to persist the focused element information when using previous/next buttons with keyboard assistance. 
This value is then used to retain focus on the previous/next buttons when the new player instance loaded on the page.

### DIFF
--- a/src/components/MediaPlayer/MediaPlayer.js
+++ b/src/components/MediaPlayer/MediaPlayer.js
@@ -44,7 +44,7 @@ const MediaPlayer = ({ enableFileDownload = false, enablePIP = false }) => {
     playlist
   } =
     manifestState;
-  const { player, currentTime } = playerState;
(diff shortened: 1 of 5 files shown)


Dananji:
Yes, my understanding was that we need to retain the focus on the next/previous buttons only when using the keyboard to improve usability. 
When using the mouse to navigate the page, I think this change doesn't add value to the user experience as much as when using the keyboard. 
Maybe this is a good question for @joncameron?

joncameron:
Good question...  I gave a quick read through WCAG and it looks like this is not required but recommended as best practice: https://www.w3.org/WAI/WCAG22/Techniques/client-side-script/SCR20.html. So it could be fine for now, but ideally both would be hitting the same event.

PR #286, opened 2023-11-14
Make tab labels focusable in the demo site

Dananji:
Re-implemented the tabs in compliance with the tabs pattern in W3C guide: https://www.w3.org/WAI/ARIA/apg/patterns/tabs/
Blog followed: https://www.w3.org/WAI/ARIA/apg/patterns/tabs/


cjcolvar (approved):
:+1:

PR #291, opened 2023-11-21
Error boundary

Dananji:
Default message;
image

Custom message (using new prop `customErrorMessage`);
image

### DIFF
--- a/src/components/MediaPlayer/MediaPlayer.js
+++ b/src/components/MediaPlayer/MediaPlayer.js
@@ -118,8 +118,8 @@ const MediaPlayer = ({ enableFileDownload = false, enablePIP = false }) => {
 
       setCIndex(canvasId);
       error ? setReady(false) : setReady(true);
-    } catch (e) {
-      showBoundary(e);
+    } catch (error) {
+      showBoundary(error);
     }
   };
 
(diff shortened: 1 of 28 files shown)


src/components/MediaPlayer/MediaPlayer.js
```diff
@@ -68,49 +74,53 @@ const MediaPlayer = ({ enableFileDownload = false, enablePIP = false }) => {
...
+      updatePlayerSrcDetails(canvas.duration, sources, canvasId, isMultiSource);
+      setIsMultiSource(isMultiSource);
 
-    setCIndex(canvasId);
-    error ? setReady(false) : setReady(true);
+      setCIndex(canvasId);
+      error ? setReady(false) : setReady(true);
+    } catch (error) {
+      showBoundary(error);
+    }
```
cjcolvar, line 123: Does it work to rename this variable here?  I think it is a bit confusing since an `error` variable already exists inside the try/catch block.
```suggestion
    } catch (e) {
      showBoundary(e);
    }
```

cjcolvar (approved):
Looks great!  There is a lot of error handling being added in this PR!  It seems like this should make ramp much more resilient.  Also nice refactorings to clean things up along the way!

PR #300, opened 2023-11-28
Setting restricted message timer when using autoplay

Dananji:
Default timeout value for message display is set to 3 seconds, this is overwritten by the `duration` property of the `placeholderCanvas` if it exists.

With autoplay turned on:

[autoplay_on.webm](https://github.com/samvera-labs/ramp/assets/1331659/ed3f51de-08d3-4a59-b309-5f81425faa83)

With autoplay turned off:

[autoplay_off.webm](https://github.com/samvera-labs/ramp/assets/1331659/d270c979-2d88-4a48-9065-3d7096329ec0)


cjcolvar (approved):
Looks good to me!

PR #305, opened 2023-12-04
Change canvas change trigger from a mutation observer to an event listener

masaball:
This PR adjusts how `canvasIndex` in the state is set. The behavior with these changes is somewhat flaky. Sometimes the hooks fire on every canvas change, sometimes they don't fire at all. I would guess it has something to do with the event listener not actually getting associated with player. 

Having the event listener and the mutation observer in place at the same time I think did provide somewhat more consistent behavior in testing when switching canvases but would trigger the transcript component to enter its loading state multiple times.

@Dananji, could you do some testing and see if you can get it working consistently?

### DIFF
--- a/src/components/Transcript/Transcript.js
+++ b/src/components/Transcript/Transcript.js
@@ -7,6 +7,7 @@ import {
   getSupplementingAnnotations,
   parseTranscriptData,
   sanitizeTranscripts,
+  getCanvasesInManifest,
   TRANSCRIPT_TYPES,
 } from '@Services/transcript-parser';
 import './Transcript.scss';
@@ -15,10 +16,10 @@ const NO_TRANSCRIPTS_MSG = 'No valid Transcript(s) found, please check again.';
 const INVALID_URL_MSG = 'Invalid URL for transcript, please check again.';
 
 /**
- * 
+ *
  * @param {String} param0 ID of the HTML element for the player on page
  * @param {Object} param1 transcripts resource
- * @returns 
+ * @returns
  */
 const Transcript = ({ playerID, manifestUrl, transcripts = [] }) => {
   const [transcriptsList, setTranscriptsList] = React.useState([]);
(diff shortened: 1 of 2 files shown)


Dananji:
The reason the transcript switching broke when there is a blank Canvas is because the `MutationObserver` we are using in the Transcript component to observe the player changes lost its connection to the player instance when we load an empty Canvas into the player without a `src` attribute.
The solution to make this work: load the hidden VideoJS player with a sample `src` attribute. This can be either a URL to an external resource or an absolute path to a local resource within Ramp. Using an external URL doesn't give Ramp the control of the resource, which could be an issue. So resolving to use a local resources requires an investigation and an implementation of Rollup configuration of how to bundle that into Ramp without ballooning the package size. 

The quick and dirty solution in this PR is to use a time interval running continuously to poll the player instance on the page, and read its `data-canvasindex` attribute to update the state in the Transcript component to load relevant transcripts into the UI.

src/components/Transcript/Transcript.js
```diff
@@ -33,14 +33,13 @@ const Transcript = ({ playerID, manifestUrl, transcripts = [] }) => {
     isMachineGen: false,
   });
   const [canvasIndex, _setCanvasIndex] = React.useState(0);
+  const [canvasIsEmpty, setCanvasIsEmpty] = React.useState(false);
```
masaball, line 36: With the logic to explicitly update this being removed does this actually do anything? Is there background processing that happens that triggers this hook?

masaball (commented):
Looks good to me, though I had a question about the `canvasIsEmpty` hook.

src/components/Transcript/Transcript.js
Dananji: No, it doesn't. Good catch!
This can be removed.

cjcolvar:
As far as I can tell this looks good to me.  With the `setInterval` polling do we not need a `src` for the hidden video.js player?

cjcolvar (approved):

PR #315, opened 2023-12-12
Retrieve and use CSRF token in Marker component

masaball:
We can retrieve the Rails CSRF token within Ramp from the HTML document. Passing that token into the parts of the component that create/handle requests, we can provide the proper headers to the `fetch` requests to get the actions to carry out as expected in Avalon.


cjcolvar (commented):
Looks great to me!  Would it be possible to add some tests to ensure the header gets added to requests when it is present in the DOM?

Dananji (approved):
Looks good :+1:

Issue #316, opened 2023-12-13
Increase Timeout Value for Inaccessible Playlist Items


joncameron:
# Description
The timeout value for moving to the next item in a playlist when an inaccessible items becomes active is 3000 milliseconds. For a better user experience this can be increased to give users more time to see that the item is not able to be viewed before the playlist moves automatically to the next item.

# Done Looks Like
- [x] Default Timeout value in Ramp increased to 10 seconds

PR #317, opened 2023-12-13
Increase Default Timeout

joncameron:
Update the timeout value for #316


cjcolvar (commented):
It looks like there are some tests that are expecting the timeout to be 3000 and need to get updated:
src/services/utility-helpers.test.js:437
src/services/utility-helpers.test.js:442
src/services/iiif-parser.test.js:257
src/services/iiif-parser.test.js:263
src/services/iiif-parser.test.js:270

cjcolvar:
I fixed these tests and rebased this branch on the latest from `main`.

PR #319, opened 2023-12-14
Primarily rely on browser for file extension assignment

### LINKED ISSUE #309: Transcript Download Doubles Extension
# Description
Downloading a Word document or text file from the Transcript component is adding an extra extension to the end of the existing full filename. 

# To Reproduce
- ramp.avalonmediasystem.org
- Use https://avalon-dev.dlib.indiana.edu/media_objects/gf06g267d/manifest.json
- Select the 2nd section
- Go to the Transcript tab and download 
- "The Legislative Organization of Congress.doc" downloads as "The Legislative Organization of Congress.doc.doc"

The original filename for supplemental files is still present in ActiveStorage in Avalon, so this could be gone back and get retrieved.

# Done Looks Like
- [ ] Storage added to retain the original filename for transcript files (or confirmed that this is the case)
- [ ] File names when downloaded are the original specified filename (except when "machine-generated" info text is inserted to the filename before the extension(s)

### LINKED ISSUE #321: Allow Download of Non-renderable Transcript Files
# Description
Transcript files which are unsupported for rendering within the Transcript component (.doc, predominantly) should still be downloadable by users, and the Download button should appear next to the transcript drop-down listing as it does for renderable files.

See https://avalon-dev.dlib.indiana.edu/media_objects/gf06g267d/section/d217qp50k for an example of the download button not appearing for a .doc file:
Screenshot 2023-12-15 at 6 28 14 AM

When VTT transcript is selected:
Screenshot 2023-12-15 at 6 33 01 AM

# Done Looks Like
- [x] Download button appears next to transcript listing for all file types

masaball:
Fixes #309, Fixes #321

This PR handles the case where a ".doc" extension was being duplicated. This approach checks for an extension in the file name before checking the URL since Avalon does not include file names in the URL. It then checks if the extension it found is part of the internal list of valid extensions. If it is, we add "(machine generated)" to the filename if necessary and then rely on the browser to assign the right extension to the download. If it is not valid, we do the same processing for "(machine generated)" and then manually add ".doc" and provide that filename to the download link. This prevents the duplication we were seeing and should prevent multiple extensions in general.

This PR also includes a fix for #321 where the download button was not rendering for unsupported file types. It was necessary for testing to have the download button always available.

~~Thinking on it more though, should we a take different approach? If the filename is "example.vtt" but the mime type is "application/msword", I think the current processing would still result in the downloaded file being "example.vtt.doc".~~

~~Rather than ever use the extension from the filename, should we instead excise anything that looks like an extension from the filename and rely on what gets generated from the mime type processing? That could be problematic though if there is a filename without extension but including periods. We could end up with weirdly truncated filenames on download if the parsing is incorrect.~~

~~Or should we approach this how @elynema suggested in the ticket, where we improve how Avalon generates the filename for the manifest and simplify/remove any processing on Ramp's side, relying on implementers to provide properly formatted "filename.ext".~~

~~Or is there another approach that could work?~~

### DIFF
--- a/src/components/Transcript/Transcript.js
+++ b/src/components/Transcript/Transcript.js
@@ -20,7 +20,7 @@ const NO_SUPPORT = 'Transcript format is not supported, please check again.';
  * @param {String} param0 ID of the HTML element for the player on page
  * @param {String} param1 manifest URL to read transcripts from
  * @param {Object} param2 transcripts resource
- * @returns 
+ * @returns
  */
 const Transcript = ({ playerID, manifestUrl, transcripts = [] }) => {
   const [transcriptsList, setTranscriptsList] = React.useState([]);
@@ -279,7 +279,7 @@ const Transcript = ({ playerID, manifestUrl, transcripts = [] }) => {
       return;
     }
 
-    // Scroll the transcript line to the center of the 
+    // Scroll the transcript line to the center of the
     // transcript component view
     autoScroll(tr, transcriptContainerRef);
   };
@@ -460,7 +460,7 @@ const Transcript = ({ playerID, manifestUrl, transcripts = [] }) => {
               selectTranscript={selectTranscript}
               transcriptData={canvasTranscripts}
               transcriptInfo={transcriptInfo}
-              noTranscript={errorMsg?.length > 0}
+              noTranscript={errorMsg?.length > 0 && errorMsg != NO_SUPPORT}
             />
           </div>
         )}
--- a/src/services/utility-helpers.js
+++ b/src/services/utility-helpers.js
@@ -181,10 +181,10 @@ export function fileDownload(fileUrl, fileName, fileExt = '', machineGenerated =
     fileNameNoExt = `${fileNameNoExt} (machine generated)`;
   }
 
-  // For Avalon style downloads, rely on the browser to properly determine file
-  // extension unless it is an unsupported format, then we provide a '.doc'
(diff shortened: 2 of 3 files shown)


cjcolvar:
@masaball Do you think it would be okay to remove the `download` attribute setting and rely on what the server sends in the `Content-Disposition` header (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition#filename)?  It looks like Avalon is handling this correctly already if you open the supplemental file url directly (https://avalon-dev.dlib.indiana.edu/master_files/d217qp50k/supplemental_files/116/transcripts).

cjcolvar:
> @masaball Do you think it would be okay to remove the `download` attribute setting and rely on what the server sends in the `Content-Disposition` header (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition#filename)? Or maybe not even use `fileDownload` at all for these links? It looks like Avalon is handling this correctly already if you open the supplemental file url directly (https://avalon-dev.dlib.indiana.edu/master_files/d217qp50k/supplemental_files/116/transcripts).

I see now that the approach I was suggesting isn't workable since avalon can't insert the `(machine generated)` on download because of limitations in activestorage.  Avalon would need to alter and store the filename which I believe we had previously ruled out.  So this leaves ramp to handle it which requires the use of `fileDownload`.

cjcolvar (approved):
Looks good as far as I can tell. :tada:

PR #322, opened 2023-12-15
Add 2 new props to set start canvas and time, and map them to start property behavior from a IIIF Manifest

Dananji:
Updated Wiki: https://github.com/samvera-labs/ramp/wiki/IIIFPlayer-Component

### DIFF
--- a/src/components/IIIFPlayer/IIIFPlayer.md
+++ b/src/components/IIIFPlayer/IIIFPlayer.md
@@ -10,7 +10,7 @@ IIIFPlayer component, provides a wrapper consisting of the Context providers con
 - `startCanvasId`: accepts a valid Canvas ID that exists within the given Manifest, this can specify the Canvas to show in Ramp on initialization. This can be mapped to the [`start` property](https://iiif.io/api/presentation/3.0/#start) in a IIIF Manifest.
 - `startCanvasTime`: accepts a valid number for a time in seconds to start playback  in the Canvas shown in Ramp on initialization.
 
-** __If the given Manifest has a `start` property defined in it, then it takes *precedence* over the `startCanvasId` and `startCanvasTime` props__
+** __`startCanvasId` and `startCanvasTime` props takes *precedence* over the `start` property in a given IIIF Manifest. Defining either prop in the IIIFPlayer component overrides the `start` property in the IIIF Manifest.__
 
 Import Ramp components individually and adjust the layout however you want. Play around with the code below.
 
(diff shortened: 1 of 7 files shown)


src/components/IIIFPlayer/IIIFPlayer.md
```diff
@@ -3,9 +3,14 @@ IIIFPlayer component, provides a wrapper consisting of the Context providers con
...
 - `manifest` : accepts a JSON object representing data in a IIIF Manifest
+
+** __Either `manifestUrl` or `manifest` is REQUIRED. If both props are given then `manifest` takes *precedence* over `manifestUrl`__
+
 - `customErrorMessage`: accepts a messagge to display to the user in the unlikely event of the component crashing, this has a default error message and it is _not required_. The message can include HTML markup.
+- `startCanvasId`: accepts a valid Canvas ID that exists within the given Manifest, this can specify the Canvas to show in Ramp on initialization. This can be mapped to the [`start` property](https://iiif.io/api/presentation/3.0/#start) in a IIIF Manifest.
+- `startCanvasTime`: accepts a valid number for a time in seconds to start playback  in the Canvas shown in Ramp on initialization.
 
-** __Either `manifestUrl` or `manifest` are REQUIRED. If both props are given then `manifest` takes *precedence* over `manifestUrl`__
+** __If the given Manifest has a `start` property defined in it, then it takes *precedence* over the `startCanvasId` and `startCanvasTime` props__
```
cjcolvar, line 13: I wonder if this should be the opposite.  I'm thinking a use case where someone wants to bookmark and open ramp at a particular canvas/time that isn't what is in the manifest (possibly via query params).  Does that make sense or is there a different way to achieve the same thing without inverting the precedence?

Dananji: My understanding was that, Ramp as a IIIF viewer always renders what's in a given IIIF Manifest. So I set the precedence this way for these, since we are introducing props/query params that can be included in the IIIF Manifest via defined properties.

And I thought we could probably add the `start` prop to Avalon manifest generation when we are loading `/media_objects/:id/section/:id` type URL in the future instead of passing the start Canvas as a prop.

If we want to change that behavior, I'm entirely okay with changing it if it provides a better handle of things.

joncameron: Let's change the behavior to have Ramp give props precedence over `start` within the manifest.

PR #323, opened 2023-12-19
Show generic error message when manifest request fails

Dananji:
image


cjcolvar (approved):
Nice!  That looks cleaner.

PR #328, opened 2024-01-02
Reduce media height

masaball:
Fixes [Avalon #5500](https://github.com/avalonmediasystem/avalon/issues/5500)

Is this the approach we want to take for resolving that issue? I think adjusting the height in this way shouldn't have knock on effects and it isn't really a noticeable reduction unless you're specifically looking for the change. But is there a better approach?


Dananji (approved):
Looks great! :100:

Issue #303, opened 2023-12-01
Change Text When No Markers Exist


joncameron:
# Description
The text users see when there are no Markers in the manifest is confusing in some contexts, like the Avalon playlist page. This text could be excised or updated to be more clear.

# Done Looks Like
- No text appears in the Markers panel when there are no Markers in a playlist manifest (only the Add Mark button display appears)

elynema:
My preference is just to have the Add Marker button display in the component with no error message if there aren't markers.

PR #333, opened 2024-01-08
Remove no markers message in MarkersDisplay component

Dananji:
Issue: https://github.com/samvera-labs/ramp/issues/303


cjcolvar (approved):
This looks good to me.  Could you please add a description to the PR or a link to an associated issue for context if we need to reference this in the future?

PR #341, opened 2024-01-16
Remove id attributes from svg elements and condense g elements

cjcolvar:
Resolves https://github.com/avalonmediasystem/avalon/issues/5565


Dananji (approved):
:+1:

PR #344, opened 2024-01-17
Add safe navigation and fallback in case there aren't any canvases in manifest

### LINKED ISSUE #339: Playlist manifest with no items is generating error message
# Description

Playlist manifest is generating with empty items[] (see https://avalon-dev.dlib.indiana.edu/playlists/39/manifest.json), but now an error is showing across all browsers (looks like it's coming from Ramp). The goal was for the playlist page to display a message "The playlist is empty" below the playlist title, instead of an error message. See details on https://github.com/avalonmediasystem/avalon/issues/5504

Firefox

Chrome

In the Ramp demo site, it gives a more generic error. 

# Questions
- [ ] Is this an issue with Ramp or an issue with Avalon? The message that was listed in Done Looks Like doesn't necessarily sound like a generic message that we would display in Ramp, since playlists are a uniquely Avalon concept.

# Done Looks Like
- [x] Message displays to user: "This playlist is empty."

cjcolvar:
Resolves #339 

This PR addresses the specific error seen in #339 and may resolve https://github.com/avalonmediasystem/avalon/issues/5504 but I don't think that it fully allows ramp to open manifests with no canvases.  When I try to open a manifest with no canvases (playlist or normal) in the ramp demo app it hits an error boundary in iiif-parser.  It looks like this happens in ramp in avalon as well but that it logs the error and moves on.

### DIFF
--- a/src/components/MetadataDisplay/MetadataDisplay.js
+++ b/src/components/MetadataDisplay/MetadataDisplay.js
@@ -82,7 +82,7 @@ const MetadataDisplay = ({
    */
   const setCanvasMetadataInState = () => {
     let canvasData = canvasesMetadataRef.current
-      .filter((m) => m.canvasindex === canvasIndex)[0].metadata;
+      .filter((m) => m.canvasindex === canvasIndex)[0]?.metadata || [];
     if (!displayTitle) {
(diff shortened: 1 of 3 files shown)


Dananji (approved):
👍

PR #348, opened 2024-01-19
Add fade out and message to indicate scrolling of structured nav

masaball:
This PR also adds a border around the structured nav. In testing, the `!important` css tags were necessary because the "ramp.css" file from the build (I think) was overriding the component specific file. Should the `!important` tags be removed because the CSS will compile into that file when we do another build?
Screenshot_20240123_104832

Screenshot_20240123_104849

### DIFF
--- a/src/components/StructuredNavigation/StructuredNavigation.js
+++ b/src/components/StructuredNavigation/StructuredNavigation.js
@@ -48,7 +48,7 @@ const StructuredNavigation = () => {
     }
   }, [manifest]);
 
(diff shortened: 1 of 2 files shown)


Dananji (approved):
👍 

As for the question of `!important` in the CSS; 
What I do to test if it works with the new CSS changes in the components, run `yarn build` locally on the branch. Then, it overwrites the `./ramp.scss` in the build files. And the demo server picks up the newly built CSS and reloads with the CSS changes in the components.
Please do not commit the build files into the branch.

Dananji (approved):
👍

PR #350, opened 2024-01-22
Read canvas summary and render as data attribute alongside label

cjcolvar:
Part of https://github.com/avalonmediasystem/avalon/issues/5581

Summary is rendered as a data attribute alongside the canvas label for playlist `ListItem`s.
In order to read the summary I had to parse the manifest again.  I'd hoped that there was a better way but couldn't find one.  Manifesto doesn't have a convenience method for getting the summary of a canvas and since it should be a language map the code to read summary is a little bit involved.

### DIFF
--- a/src/services/iiif-parser.js
+++ b/src/services/iiif-parser.js
@@ -1,4 +1,4 @@
-import { parseManifest } from 'manifesto.js';
+import { parseManifest, PropertyValue } from 'manifesto.js';
 import mimeDb from 'mime-db';
 import sanitizeHtml from 'sanitize-html';
 import {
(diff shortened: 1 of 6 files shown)


src/services/iiif-parser.js
```diff
@@ -580,9 +581,17 @@ export function getStructureRanges(manifest) {
       if (isCanvas && canvasInfo.range != undefined) {
         duration = canvasInfo.range.end - canvasInfo.range.start;
       }
+      if (isCanvas) {
+        let parsedManifest = parseManifest(manifest);
+        let summaryProperty = parsedManifest.getSequences()[0].getCanvasById(getCanvasId(canvases[0]))?.getProperty('summary');
```
masaball, line 586: Instead of importing `getSequences` from manifest.js and parsing the manifest here, I think we can just use the `parseSequences` function from the utility-helpers file which is already imported.
```suggestion
        let summaryProperty = parseSequences(manifest)[0].getCanvasById(getCanvasId(canvases[0]))?.getProperty('summary');
```

masaball (approved):

Issue #325, opened 2023-12-21
IIIF caption/subtitle cookbook recipe no longer works in Ramp


elynema:
# Description

The IIIF cookbook has a recipe for providing a caption/transcript file in a supplementing annotation as a vtt file.

https://iiif.io/api/cookbook/recipe/0219-using-caption-file/

Loading this in the Ramp demo site, the vtt file no longer loads to display as transcripts and captions in Ramp. https://ramp.avalonmediasystem.org/?iiif-content=https://iiif.io/api/cookbook/recipe/0219-using-caption-file/manifest.json

Could this be related to our work to have Ramp recognize captions and transcripts specifically coming out of Avalon using the id of the annotation body?
https://github.com/avalonmediasystem/avalon/pull/5226

elynema:
Re-test captions in playlists, media items, and mobile devices again, as well as with cookbook examples. This involved a change to how we prevent duplication of captions between HLS manifest and IIIF manifest captions.

joncameron:
Captions aren't loading on the Ramp demo page for me on iOS for https://ramp.avalonmediasystem.org/?iiif-content=https://iiif.io/api/cookbook/recipe/0219-using-caption-file/manifest.json. They don't appear within the player, either before playback or in full-screen/native player mode.
@Dananji 

Screenshot of console:
<img width="532" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL3dDQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--728c7ed83c08718a79e3753d8d4622810f005929/Screenshot%202024-01-26%20at%2011.54.16%E2%80%AFAM.png" alt="Screenshot 2024-01-26 at 11.54.16 AM.png" />

elynema:
Both transcripts and captions are loading properly in Chrome.

In Safari, the transcript loads, but there is no CC icon in the player. I see the same console error that Jon posted above.

In Firefox, the captions load, but the transcript component is empty. The console seems to think it is a permissions issue. Not sure if this is unique to my browser configuration or happening for everyone.
`Security Error: Content at https://ramp.avalonmediasystem.org/?iiif-content=https://iiif.io/api/cookbook/recipe/0219-using-caption-file/manifest.json may not load data from https://fixtures.iiif.io/video/indiana/lunchroom_manners/lunchroom_manners.vtt.`

[...]

PR #353, opened 2024-01-22
Filter duplicated captions instead of switching off native text track behavior in VideoJS

Dananji:
With Avalon IIIF Manifests, Ramp receives identical captions/subtitles information from both IIIF Manifest and HLS manifest. When VideoJS in Ramp is using its normal text track handling with these IIIF Manifests, captions/subtitles get duplicated in the player's captions/subtitles menu because it uses captions/subtitles information in both manifests. 

Using VideoJS option `html5: { nativeTextTracks: true }` prevents captions/subtitles duplication for Avalon IIIF Manifests, but this hides text tracks for other IIIF Manifests where there are no HLS manifests with captions/subtitles information.

Therefore in this PR; `html5: { nativeTextTracks: true }` is removed and added a filter in the JS code to programmatically remove the duplicated captions/subtitles in the player's captions/subtitles menu.

Related issue: #325

### DIFF
--- a/src/components/MediaPlayer/MediaPlayer.js
+++ b/src/components/MediaPlayer/MediaPlayer.js
@@ -326,10 +326,6 @@ const MediaPlayer = ({ enableFileDownload = false, enablePIP = false }) => {
       ? playerConfig.sources[srcIndex]
       : playerConfig.sources,
     tracks: playerConfig.tracks,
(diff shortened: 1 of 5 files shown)


cjcolvar (approved):
Looks good to me!

PR #360, opened 2024-01-24
Display transcript content on first load

Dananji:
Related issue: https://github.com/avalonmediasystem/avalon/issues/5584


masaball (approved):
Looks good as far as I can tell

PR #362, opened 2024-01-26
Improve handling of download filenames in Transcript and Supplemental File components

### LINKED ISSUE #309: Transcript Download Doubles Extension

masaball:
Resolves #309

### DIFF
--- a/src/components/SupplementalFiles/SupplementalFiles.js
+++ b/src/components/SupplementalFiles/SupplementalFiles.js
@@ -44,7 +44,7 @@ const SupplementalFiles = ({ itemHeading = "Item files", sectionHeading = "Secti
 
   const handleDownload = (event, file) => {
     event.preventDefault();
-    fileDownload(file.id, file.filename, file.fileExt);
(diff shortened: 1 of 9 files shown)


Dananji (approved):
👍 
Great work on documenting these changes via comments!

PR #364, opened 2024-01-26
Add padding to playlist structure nav

Dananji:
_Playlist structure_

Before:
Before

After:
After

And this doesn't change the regular structure display.

Related issue: https://github.com/avalonmediasystem/avalon/issues/5589


cjcolvar (approved):

PR #367, opened 2024-01-30
Override Safari's native text track functionality

Dananji:
Related issue: https://github.com/samvera-labs/ramp/issues/325


masaball (approved):
Looks good

PR #369, opened 2024-01-31
Use touch event handlers in progress bar for mobile/tablet devices

Dananji:


src/services/browser.js
```diff
@@ -0,0 +1,279 @@
...
+ */
+export let IS_ANDROID = false;
+
+/**
+ * The detected Android version - or `null` if not Android or indeterminable.
+ *
+ * @static
+ * @type {number|string|null}
+ */
+export let ANDROID_VERSION;
```
cjcolvar, line 33: For consistency should this be initialized to `null`?
```suggestion
export let ANDROID_VERSION = null;
```

cjcolvar (commented):
I like the autoscroll checkbox and think that code looks great!
The changes for mobile/tablet playback seem fairly big given all the changes to events.  Would it be hard to separate these into two different PRs?  I'm thinking that if we merge them separate it will be easier to isolate or revert the playback changes if needed.

src/services/browser.js
Dananji: Absolutely! 💯

Dananji:
Did you mean split the transcript and event handlers into 2 separate PRs? I don't think it would be hard to do. Let me do that real quick.

Dananji:
New PR with transcript changes: https://github.com/samvera-labs/ramp/pull/371

Issue #354, opened 2024-01-23
[BUG] Media player issues in Chrome and Safari on iPad


elynema:
# Description

iPads do not default to opening the native player like iPhones do.

In Safari on the iPad, we are seeing the same player we do in Safari on a computer, both in non-full screen and full screen mode. It is being non-responsive / very laggy when attempting to scrub through the progress bar in particular. That could potentially be resolved by fixing the issue we are seeing in Safari on computer in Ramp #338. samvera-labs/ramp#338

In Chrome on an iPad, it will play non-full screen in Chrome with the video.js player, but if you open full screen it appears to try to leverage the native player. Unfortunately, the scrubbing and other buttons in the non-full screen player in Chrome on the iPad are also showing huge latency / unresponsiveness. After a few attempts to scrub, the media player controls sometimes disappear and then don't come back for many seconds when I click on the video. The native player on iPad is also unresponsive when attempting to scrub, jump forward/back. If I attempt to scrub, then the player stops responding to play/pause, etc. The current time in the progress bar seems to get stuck, even when video is playing.

The non-responsiveness is also happening in li'l scrub, as well.

I attempted a video of the issues in Chrome on an iPad,  but it's probably best to experiment with it directly in Browserstack.

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOVB1QWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--50d74d229ee8c60f43c64751f0e256f9085e2a80/Screen%20Recording%202024-01-23%20at%208.59.10%20AM.mov)

# Done Looks Like
- [x] Users can scrub using the progress bar in both Safari and Chrome on iPad in non-full screen mode
- [x] Users can scrub using the progress bar in both Safari and Chrome on iPad in full screen mode
- [x] Users can use the skip forward / back buttons in the native player in Chrome on iPad in full screen mode
- [x] The current time playhead advances properly in both Safari and Chrome even with scrubbing
- [x] Users can scrub using li'l scrub in Safari and Chrome on iPad in non-full screen mode
- [x] Transcript follows scrubbing in the progress bar in Safari on iPad in non-full screen mode

joncameron:
The timeupdate event has been buggy in some contexts, and sometimes stops working in the browser. This could be the case here since all of the affected events are being set by timeupdate. Could be something in the code there which is breaking.

[...]

PR #371, opened 2024-02-02
Replace pointer handlers for transcript auto-scroll check with a checkbox

Dananji:
Introduce a checkbox to turn on/off auto-scroll for transcript content:

With time-synced transcripts:
Screenshot 2024-02-01 at 3 57 02 PM

With non time-synced transcripts:
Screenshot 2024-02-01 at 4 03 15 PM

Related issue: #354


elynema:
Only initially tested in Ramp demo site in Chrome on Mac and Safari/Chrome on iPad. In those scenarios, this is an elegant solution. I tried turning the auto-scroll on/off, scrubbing through media via progress bar and transcript, letting the transcript active cue progress below the visible display and turning auto-scroll back on.

Auto-scroll is on by default and most users will just leave it that way and it will continue to work well. Users who have an issue in a scenario like a mobile device can turn it off if needed.

PR #378, opened 2024-02-05
Disable volume slider for mobile devices

### LINKED ISSUE #352: [BUG] Volume controls not working in media player on Android devices
# Description

Volume cannot be adjusted in non-full screen or full-screen player mode on Android devices. It just toggles back and forth between muted and un-muted when you click on it to try to open the slider. This is only on video; audio volume can be adjusted because it is a static control. Use user agent or other signifying value to change configuration in Video.js for iOS and Android devices

Tested with Samsung Galaxy on both Chrome and Firefox.

We'll check what exists in the Avalon 7.6 code for handling these cases and implement in 7.7 as appropriate. 

# Done Looks Like
- [x] Volume slider is disabled/does not appear in the media player for video content in iOS and Android (phone and tablet) devices/browsers
- [x] Volume button functions as a mute/unmute button (no volume slider) for video in iOS and Android (phone and tablet) devices/browsers
- [x] Change user agent or other signifying value check in Video.js configuration to suppress volume slider

masaball:
Phones and tablets lack the ability to :hover, so the behavior of the volume slider is inconsistent and confusing on such devices. We can provide consistent behavior by disabling the slider and only providing a toggle for mute/unmute.

Resolves #352

### DIFF
--- a/src/components/MediaPlayer/MediaPlayer.js
+++ b/src/components/MediaPlayer/MediaPlayer.js
@@ -304,7 +304,7 @@ const MediaPlayer = ({ enableFileDownload = false, enablePIP = false }) => {
         'durationDisplay',
         hasStructure ? 'videoJSTrackScrubber' : '',
         playerConfig.tracks.length > 0 ? 'subsCapsButton' : '',
-        'volumePanel',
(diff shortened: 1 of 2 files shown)


cjcolvar (approved):
Looks great!

PR #387, opened 2024-02-06
Show track scrubber in playlist context

Dananji:
Related issue: https://github.com/avalonmediasystem/avalon/issues/5624

### DIFF
--- a/src/components/MediaPlayer/MediaPlayer.test.js
+++ b/src/components/MediaPlayer/MediaPlayer.test.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { act, render, screen } from '@testing-library/react';
+import { act, render, screen, waitFor } from '@testing-library/react';
 import { withManifestAndPlayerProvider } from '../../services/testing-helpers';
 import MediaPlayer from './MediaPlayer';
 import { ErrorBoundary } from 'react-error-boundary';
@@ -22,7 +22,7 @@ describe('MediaPlayer component', () => {
     console.error = originalError;
   });
 
-  describe('with audio manifest', () => {
+  describe('with a regular audio Manifest', () => {
     beforeEach(() => {
       const PlayerWithManifest = withManifestAndPlayerProvider(MediaPlayer, {
         initialManifestState: { ...manifestState, manifest: audioManifest, canvasIndex: 0 },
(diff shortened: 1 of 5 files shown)


src/components/MediaPlayer/VideoJS/components/js/VideoJSTrackScrubber.js
```diff
@@ -201,19 +208,25 @@ function TrackScrubberButton({ player, trackScrubberRef, timeToolRef }) {
 	 */
 	const timeUpdateHandler = () => {
 		if (player.isDisposed()) return;
```
masaball, line 210: I did not think to test the track scrubber when I was looking at the issue with the flashing struct nav. Just to be safe that we don't accidentally reintroduce that behavior, we should probably have this time update return on ended as well.
```suggestion
		if (player.isDisposed() || player.ended()) return;
```

src/components/MediaPlayer/VideoJS/components/js/VideoJSTrackScrubber.js
```diff
@@ -24,15 +25,19 @@ class VideoJSTrackScrubber extends vjsComponent {
...
-		player.ready(() => {
-			this.mount();
-		});
+		/* When player is ready and the trackScrubber element is initialized,
+		call method to mount React component.
+		*/
+		if (this.options.trackScrubberRef.current) {
+			player.ready(() => {
+				this.mount();
+			});
```
masaball, line 34: It looks like everything in this file is indented at least twice as far as it should be for some reason?

src/components/MediaPlayer/VideoJS/components/styles/VideoJSTrackScrubber.scss
```diff
@@ -15,11 +15,19 @@
...
 	justify-content: space-between;
 
+	@media (max-width: 680px) {
+		width: 99.65%;
+	}
+
+	@media (min-width: 1081px) {
+		width: 99.85%;
+	}
+
```
masaball, line 30: This file also has double width indents.

src/components/MediaPlayer/MediaPlayer.test.js
```diff
@@ -46,7 +46,7 @@ describe('MediaPlayer component', () => {
     });
   });
 
-  describe('with video manifest', () => {
+  describe('with a regualr video Manifest', () => {
```
masaball, line 49:
```suggestion
  describe('with a regular video Manifest', () => {
```

masaball (commented):
Looks good for the most part. Have one suggestion that may prevent a regression and then some styling questions.

src/components/MediaPlayer/VideoJS/components/js/VideoJSTrackScrubber.js
Dananji: It was formatting issue when I created the files. Thanks for catching that. I wasn't seeing that in my editor with its settings.

masaball (approved):

PR #389, opened 2024-02-07
Disable time updates when media has ended

masaball:
Switching our time updates to trigger on an interval caused the handler to be called continuously, even after media had ended. This resulted in undesirable behavior in the structured nav when the last item would finish. Preventing the handler from firing when the player is in the 'ended' state fixes the behavior.


Dananji (approved):
👍

PR #392, opened 2024-02-07
Remove z-index for player control bar in mobile devices, add an additional check for inactivity

Dananji:
The `IS_IPAD` check is flaky when setting the VideoJS options, therefore adding an additional check.

Related issue: https://github.com/avalonmediasystem/avalon/issues/5633

### DIFF
--- a/src/components/MediaPlayer/MediaPlayer.js
+++ b/src/components/MediaPlayer/MediaPlayer.js
@@ -13,7 +13,7 @@ import {
 } from '../../context/player-context';
 import { useErrorBoundary } from "react-error-boundary";
 import './MediaPlayer.scss';
-import { IS_MOBILE, IS_IPAD, IS_SAFARI } from '@Services/browser';
+import { IS_MOBILE, IS_IPAD, IS_SAFARI, IS_TOUCH_ONLY } from '@Services/browser';
 
 const PLAYER_ID = "iiif-media-player";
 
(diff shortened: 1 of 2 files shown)


src/components/MediaPlayer/MediaPlayer.js
```diff
@@ -288,7 +288,7 @@ const MediaPlayer = ({ enableFileDownload = false, enablePIP = false }) => {
     // Setting inactivity timeout to zero in mobile and tablet devices translates to
     // user is always active. And the control bar is not hidden when user is active.
     // With this user can always use the controls when the media is playing.
-    inactivityTimeout: (IS_MOBILE || IS_IPAD) ? 0 : 2000,
+    inactivityTimeout: (IS_MOBILE || IS_TOUCH_ONLY || IS_IPAD) ? 0 : 2000,
```
masaball, line 291: It is interesting that we would need all these checks. `IS_TOUCH_ONLY` and `IS_IPAD` (as part of `IS_IOS`) are part of the `IS_MOBILE` assignment logic, so I would expect any situation that puts those as true to put `IS_MOBILE` as true. Maybe we have a bug in the `IS_MOBILE` assignment somewhere?

masaball (approved):
Looks good!

src/components/MediaPlayer/MediaPlayer.js
Dananji: All the check seem to be in order to me. 

I tested again only with `IS_TOUCH_ONLY` and it's working. Could be I missed something when I tested before. I'll change this and commit again.

masaball (approved):

PR #393, opened 2024-02-07
Add key to track map in VideoJS initializer

masaball:
We were getting a console error about list children requiring a key. This was related to our caption tracks array. I think we were seeing this because [JSX elements in a `map()`](https://react.dev/learn/rendering-lists#keeping-list-items-in-order-with-key) always require keys. Adding a key attribute to the track element within the map in VideoJS initialization resolves the error.


Dananji (approved):
👍

PR #395, opened 2024-02-07
Use localStorage to make volume and quality settings sticky

### LINKED ISSUE #390: Volume and Quality Selector Resets to Default On Each Canvas Change
# Description

cjcolvar:
Resolves #390 

This PR adds a helper service to make it easy to use localStorage as a React hook.  (It is adapted from https://blog.logrocket.com/using-localstorage-react-hooks/)  The `useLocalStorage` then gets used to store and retrieve volume and quality selections.  Volume is then set on player ready.  Quality is set by adjusting the `selected` attribute on sources right before instantiating the videojs player.  I thought keeping the storing/retrieving/setting all in the VideoJSPlayer component was simpler and kept it contained instead of trying to change the parser code to include the quality selection from localStorage.

I tried to write tests for the helper service but wasn't able to figure out how to setup the test.  I also tried adding tests of the rendered VideoJSPlayer component to check that setting volume or quality wrote the setting into the localStorage and that a setting in localStorage would be restored when the player rendered.  A couple problems I ran into with the VideoJSPlayer component testing were the player not getting to the ready state so not emitting events like `ready`, `volumechanged`, and `qualityRequested` and the `startQuality` value not getting set in the localStorage mock hash for an unknown reason (it was always present but null :shrug:).  I have committed all of these tests and marked them as skip so we can try to make them work in the future.  @Dananji suggested that it could be that the approach to testing the `VideoJSPlayer` component needs to change to be at a lower level: mock VideoJS entirely and simulate the events being emitted so we don't need to rely on the player initializing and getting to the ready state.

### DIFF
--- a/src/components/MediaPlayer/MediaPlayer.test.js
+++ b/src/components/MediaPlayer/MediaPlayer.test.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { act, render, screen } from '@testing-library/react';
+import { act, render, screen, waitFor } from '@testing-library/react';
 import { withManifestAndPlayerProvider } from '../../services/testing-helpers';
 import MediaPlayer from './MediaPlayer';
 import { ErrorBoundary } from 'react-error-boundary';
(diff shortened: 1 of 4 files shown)


Dananji (approved):
👍

PR #397, opened 2024-02-08
Assign id value to empty VideoJS instance

masaball:
Empty VideoJS instances were not being initialized with the 'iiif-media-player' id attribute. This broke some functionality in Avalon and prevented the proper CSS rules from being applied. This commit also adds a z-index to the empty canvas message so that if the player CSS breaks again, the message should still be layered on top of it.


Dananji (approved):
👍

PR #398, opened 2024-02-08
Update autoadvancetoggle component styleguidist docs

Dananji:


src/components/AutoAdvanceToggle/AutoAdvanceToggle.md
```diff
@@ -1,4 +1,6 @@
-AutoAdvanceToggle description
+AutoAdvanceToggle component provides the ability to turn on/off auto-advance from a Canvas to the next when viewing a IIIF Manifest. It reads the `behavior` property at the Manifest-level, and then parses and stores the given values to a Boolean flag within Ramp's state management.
+
+When `behavior` is not specified in the Manifest, this is set to not auto-advance as the default temporal behavior if not specified is `no-auto-advance` per the [IIIF Presentation 3.0 specification](https://iiif.io/api/presentation/3.0/#behavior).
```
cjcolvar, line 3: What do you think about this slight wording change?
```suggestion
When `behavior` is not specified in the Manifest, this is set to not auto-advance as the default temporal behavior due to not specified being the same as `no-auto-advance` per the [IIIF Presentation 3.0 specification](https://iiif.io/api/presentation/3.0/#behavior).
```

cjcolvar (commented):
This is great!

PR #409, opened 2024-02-12
Fix broken hotkeys when switching Canvas and for mute in Chrome

Dananji:
Related issue: https://github.com/avalonmediasystem/avalon/issues/5645

### DIFF
--- a/src/services/utility-helpers.js
+++ b/src/services/utility-helpers.js
@@ -497,9 +497,8 @@ export function autoScroll(currentItem, containerRef) {
  * @param {String} id player instance ID in VideoJS
  * @returns 
  */
-export function playerHotKeys(event, id) {
-  let player = document.getElementById(id);
-  let playerInst = player?.player;
+export function playerHotKeys(event, player) {
+  let playerInst = player?.player_;
 
   let inputs = ['input', 'textarea'];
   let activeElement = document.activeElement;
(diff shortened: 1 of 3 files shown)


src/services/utility-helpers.js
```diff
@@ -497,9 +497,8 @@ export function autoScroll(currentItem, containerRef) {
  * @param {String} id player instance ID in VideoJS
  * @returns 
  */
-export function playerHotKeys(event, id) {
-  let player = document.getElementById(id);
-  let playerInst = player?.player;
+export function playerHotKeys(event, player) {
+  let playerInst = player?.player_;
```
masaball, line 501: I have not seen `player.player_` used before. Is there a difference from `player.player`? Is it just what's accessible in this context?

masaball (approved):
Looks good, though have a question on the `playerInst` variable assignment.

src/services/utility-helpers.js
Dananji: I think it's because we are accessing the Video.js instance's property instead of the player object from the DOM element like before. I just checked and it seems it works with `player.player()`, which accesses the player API within Video.js. Do you want me to switch it out to that, since it seems more robust?

masaball: Yeah, if it works and is more robust/supported I think we should switch over to it.

masaball (approved):

PR #411, opened 2024-02-13
Force text track font size in mobile contexts

masaball:
Text tracks were generating with an exceedingly large font in full screen portrait mode on mobile browsers. Overriding the native VideoJS styling and explicitly defining a font size remedies the issue.

### DIFF
--- a/src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
+++ b/src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
@@ -20,7 +20,7 @@ import {
   getCanvasIndex,
 } from '@Services/iiif-parser';
 import { checkSrcRange, getMediaFragment, playerHotKeys } from '@Services/utility-helpers';
-import { IS_IPAD, IS_MOBILE } from '@Services/browser';
+import { IS_ANDROID, IS_IPAD, IS_MOBILE, IS_TOUCH_ONLY } from '@Services/browser';
 import { useLocalStorage } from '@Services/local-storage';
 
 /** VideoJS custom components */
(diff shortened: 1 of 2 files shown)


src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
```diff
@@ -584,6 +584,18 @@ function VideoJSPlayer({
...
+  // Not all Android tablets return 'Android' in the useragent, so assume touch 
+  // devices are tablets. This should be safe because iPhones use the native player.
+  let videoClass = '';
+  if (IS_ANDROID) { 
+    videoClass = "video-js vjs-big-play-centered android";
+  } else if (IS_TOUCH_ONLY) {
+    videoClass = "video-js vjs-big-play-centered tablet"
+  } else { 
+    videoClass = "video-js vjs-big-play-centered"; 
+  };
```
masaball, line 597: Thinking on it more, would it be better to do something like `else if (IS_TOUCH_ONLY && !IS_IOS) {...} else if (IS_IPAD) {...}` to explicitly exclude iPhones and iPods? Or should we not worry about it because of the native player overriding VideoJS on those devices?

cjcolvar: Might be good practice just to make sure in case we add more to those classes later.

src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
```diff
@@ -584,6 +584,21 @@ function VideoJSPlayer({
     return null;
   };
 
+  // Classes for setting caption size based on device
+  let videoClass = '';
+  console.log(IS_IPAD);
```
cjcolvar, line 589: It looks like this was left in from debugging.  Can you please remove this?

cjcolvar (approved):
:+1:

Issue #400, opened 2024-02-09
[BUG] Scrubbing crashes full screen mode on iPad in iOS 17


elynema:
# Description

Scrubbing after entering full-screen mode on iPad in iOS 17 crashes full-screen mode. After that clicking on full-screen mode flashes open full-screen, but it immediately closes again. This happens on both media object and playlist pages.

This does not happen in Chrome on iOS 17 or in earlier versions of iOS. Note that Safari on iPad does *not* make use of the native player; it uses the Ramp player. It's unlikely that this issue is replicated on iPhones, which do use the native player. Chrome on iPad also uses the native player.

To reproduce:
1. Start playback of a video item in Safari on iPad iOS 17
2. Enter full screen mode
3. Attempt to scrub forward by clicking or dragging
4. After 1 or 2 clicks, full-screen will crash and return to the media object view page
5. Attempting to enter full screen again will cause it to crash again

<img width="543" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMDlJQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--8247e7184410816f1073ac8bf4b57cb1573a8418/Screen%20Shot%202024-02-09%20at%201.04.03%20PM.png" alt="Screen Shot 2024-02-09 at 1.04.03 PM.png" />

joncameron:
Tested in Browserstack on iOS 17 via iPad and could no longer reproduce.

PR #412, opened 2024-02-13
Fix fullscreen crashing in Safari in iPad on iOS 17

Dananji:
Related issue: https://github.com/samvera-labs/ramp/issues/400


cjcolvar (approved):

PR #414, opened 2024-02-14
Make muted be a sticky setting

cjcolvar:
The volume sticky setting doesn't track when a user mutes/unmutes.  This PR makes muted be a sticky setting alongside the volume sticky setting.

Note there is not a specific event for muted but when a user mutes/unmutes the `volumechange` event is triggered.

#390


Dananji (approved):
👍

PR #417, opened 2024-02-15
Use safe navigator when setting quality

masaball:
The `sources` variable is not set when there is a blank canvas, causing an error. This breaks generation of the dummied out player instance that we rely on for certain actions within Avalon. Using the safe navigator prevents the error, allowing generation of the player to finish.


Dananji (approved):
👍

Issue #346, opened 2024-01-18
Undefined Error for Manifest with No Items


joncameron:
# Description
An example manifest from Tanya Clement that contains no items is resulting in a fatal error in Ramp when loaded.

# To Reproduce
- Go to ramp.avalonmediasystem.org
- Load https://tanyaclement.github.io/harvard1953/august-3-evening-part-one/manifest.json
- `o.items is undefined` error appears

# Done Looks Like
- [x] Information in the manifest is populated to the corresponding components when there is no item present in the manifest

joncameron:
I now see the "This item cannot be played." message instead of the error on the Ramp demo site.

PR #426, opened 2024-02-21
Test audiannotate

Dananji:
Related issue: https://github.com/samvera-labs/ramp/issues/346

Changes in the parser:
- Fail gracefully for empty canvases when represented with a body but no URIs in it
- Fix condition for reading `items` list under `AnnotationPage` property, handle cases when there are no `items` list

These changes are related to reading AudiAnnotate manifests in Ramp.


masaball (approved):
Looks good!

Issue #379, opened 2024-02-05
[BUG] Null value in metadata prevents manifests from loading in Ramp


elynema:
# Description

Some manifests from MCO (pre-7.7) do not render in Ramp. There is a console error related to metadata property. Note that prior to 7.7, metadata for the item was not really included in the manifest.

```
iiif-parser.js:516 iiif-parser -> getMetadata() -> cannot parse manifest,  TypeError: Cannot read properties of null (reading 'replace')
    at iiif-parser.js:532:22
    at Array.map (<anonymous>)
    at pe (iiif-parser.js:530:14)
    at iiif-parser.js:513:36
    at MetadataDisplay.js:52:30
    at Lc (react-dom.production.min.js:262:359)
    at t.unstable_runWithPriority (scheduler.production.min.js:18:343)
    at Hr (react-dom.production.min.js:122:325)
    at Ic (react-dom.production.min.js:261:308)
    at react-dom.production.min.js:261:215
    at L (scheduler.production.min.js:16:224)
    at A.port1.onmessage (scheduler.production.min.js:12:346)
```

It appears to be the null value for the Note field is causing the problem, as manifests without this null value render in Ramp.

<img width="754" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMzByQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--2a5f579442314ccd2c38399047cf4ef42eec840a/image.png" alt="image.png" />

Nearly all of the Indiana School of the Sky recordings have this null value, but there are other records that do as well.
School of the Sky: https://media.dlib.indiana.edu/catalog?f%5Bcollection_ssim%5D%5B%5D=Indiana+School+of+the+Sky&q=&search_field=all_fields

Other example: https://media.dlib.indiana.edu/media_objects/tm70nc75d

These manifests do work in Clover and Aviary, so it seems we might want to be able to handle a null case in Ramp.

I've attached a sample manifest as this null value *may* be removed when we upgrade to 7.7 (I'm not sure if it is coming from the actual metadata somehow or from the way the manifest is generated).

[...]

PR #427, opened 2024-02-21
Interpret null values in metadata as empty strings

Dananji:
With this change metadata with null values would show up in the MetadataDisplay component as below;

Screenshot 2024-02-21 at 4 03 53 PM
Screenshot 2024-02-21 at 4 03 21 PM

This is identical to the behavior in both Clover and UV.

@joncameron Does this look okay? Or do you think removing the metadata field in the display works better?

Related issue: https://github.com/samvera-labs/ramp/issues/379


joncameron:
@Dananji This looks OK to me, especially since it matches other players' behavior. It looks a little odd to have a field with no value but if it's in the manifest it makes sense to render out even with a value of nil/null/empty string.

masaball (approved):
:+1:

PR #428, opened 2024-02-22
Change caption filtering to always use VideoJS' tracks instead of HLS tracks

Dananji:
Related issue: https://github.com/avalonmediasystem/avalon/issues/5655

This PR changes the captions filtering in the check, to always use the tracks injected by Video.js instead of using HLS tracks. 
Reason for using Video.js tracks: sometimes the HLS' tracks can have `DEFAULT=NO` which sets the captions off by default. Video.js tracks are always created with `default` set to `true` in the Ramp code, ensuring that captions are on by default.


cjcolvar (approved):

PR #432, opened 2024-02-23
Fix auto advance and player crashing on Canvas change

Dananji:
Fixes in the PR:
- Auto advance not working when the first Canvas in the Manifest is empty
- Video.js instance crashing when switching between canvases due to the `player.paused()` check referring to a null instance for a fraction of a second


src/components/MediaPlayer/VideoJS/components/js/VideoJSProgress.js
```diff
@@ -117,7 +117,7 @@ class VideoJSProgress extends vjsComponent {
     // Some items, particularly in playlists, were not having `player.ended()` properly
     // set by the 'ended' event. Providing a fallback check that the player is already
     // paused prevents undesirable behavior from excess state changes after play ending.
-    if (curTime >= end && !player.paused()) {
+    if (curTime >= end && player && !player.paused()) {
```
cjcolvar, line 120: Good catch!

cjcolvar (commented):
Looks good.

cjcolvar (approved):

Issue #347, opened 2024-01-19
[BUG] placeholderCanvas not displaying for inaccessible items in playlists on iOS devices


elynema:
# Description

When viewing inaccessible (or lock icon) items in a playlist on an iOS device in either Safari or Chrome, the black rectangle with placeholderCanvas text does not appear. Instead, the player simply disappears and only the playlist item title shows with markers / source item details below.

This happens on the Ramp demo site, as well, so suspect it is a Ramp issue.

Tested on iPhone 11 in Browserstack.

I have not tested inaccessible / items on the media player page to see if the same thing happens there.

<img width="412" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN2pnQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f453659e65bc06972bde03a13622e50d81503b08/bs_realios_Mobile_iPhone%2011-13.0.jpg" alt="bs_realios_Mobile_iPhone 11-13.0.jpg" />

Dananji:
This can be tested on Ramp demo site.

joncameron:
Testing on the Ramp site with https://avalon-dev.dlib.indiana.edu/playlists/94/manifest.json, I see the empty player but not the "You do not have permission to playback this item." message for an inaccessible item. Looks like that div with the text is present on the page but not displaying inside the blank player area. 

@Dananji is this something that could be fixed with a styling change?

Dananji:
The new styling changes worked in the PR were not reflected in the demo site, since the CSS files were not re-built on the Netlify deploy.
I built the CSS files manually and pushed them to `main`, so the demo site now uses the new CSS changes in the PR.

PR #435, opened 2024-02-23
Use alternate CSS properties to set the Canvas message height

Dananji:
In iOS versions < 14 `aspect-ratio` CSS property is not supported, and this caused the placeholder canvas message to not to display in Safari browsers.

Related issue: https://github.com/samvera-labs/ramp/issues/347


cjcolvar (approved):
Looks tricky!  Nice work!

Issue #422, opened 2024-02-16
[BUG] Auto-play does not advance for audio items on iOS


elynema:
# Description

Auto-play does not reliably advance for audio only items both on the media item page and in playlists when the end of the currently playing canvas is reached.

On iPhone (Safari), the next canvas does not begin playing automatically in either playlists or media items when transitioning:

- audio canvas to audio canvas
- audio canvas to video canvas
- video canvas to audio canvas

Video to video canvas transitions do continue to auto-play.

On iPad (Safari), the next canvas does not begin playing automatically when transitioning:

- audio canvas to audio canvas (media item page)
- audio canvas to video canvas (media item page). This works inconsistently on the playlist page.

Other transitions do seem to auto-play fine on iPad.

Testing item with multiple audio sections: https://avalon-dev.dlib.indiana.edu/media_objects/fn106x94r
Testing item with both video and an audio section: https://avalon-dev.dlib.indiana.edu/media_objects/g158bh28p
Testing playlist with video and multiple audio sections: https://avalon-dev.dlib.indiana.edu/playlists/56

joncameron:
Play/pause status of player in state management may be inconsistent in iOS, causing this behavior (could be something else!).

masaball:
This seems to be a consequence of the way that Apple devices handle [video/audio autoplay](https://webkit.org/blog/6784/new-video-policies-for-ios/) in that they do everything they can to prevent it unless the media has no audio or is muted. They do this as a measure to save on website bandwidth and device battery use. I have not been able to find any more recent discussion from Apple or Webkit and the behavior observed in investigating this issue was overall consistent with the behavior described in the linked blog post.

When the player advances to an audio item, iOS detects the attempt to play and cancels the event via a rejected Promise. This sometimes also occurs with video autoplay but it seems dependent on iOS version and prior player state. I experimented with a few different options for trying to emulate a user action by "clicking" the play button via javascript instead of using the builtin `player.play()` or `video.play()` functions from Video.js/HTML video element. Nothing worked in my testing, though. 

I am unsure why videos are autoplaying on some versions of iOS (it was most consistent in iOS 17 in testing). It likely has something to do with videos using the native player rather than an inline player but I was unable to find any discussion about this or anything obvious in our code that might otherwise explain the behavior. It is also possible that there could be device settings that adjust the behavior and BrowserStack (inadvertently?) has different settings enabled on different devices.

[...]

PR #444, opened 2024-03-06
Prevent iOS rejected promise from raising error boundary

masaball:
In investigation for #422 I was unable to find a way to force autoplay of all items because iOS locks down that functionality as tight as they can (I am unsure why autoplay of videos works on some devices, probably something to do with them using the native player). However, I did find that certain versions of iOS triggered a runtime error within Ramp by the method of how they prevent/cancel the play event. This error catch should prevent that error from triggering the error boundary and allow users to interact regularly with the Ramp components.


Dananji (approved):
Good catch 👍 

I have one suggestion, could we maybe add the explanation in the PR to the code as a comment? The code itself might seem like incomplete error handling without the proper explanation.

Dananji (approved):
🎉

PR #445, opened 2024-03-08
Omit active tab navigation with left/right key events in hotkeys

Dananji:
Related issue: https://github.com/avalonmediasystem/avalon/issues/5666

Trigger player hotkeys when focus is not on an input, textarea or on a navigation tab with the key pressed is one of left/right arrow keys. This specific combination of keys with a focused navigation tab is avoided to allow keyboard navigation between tabbed UI components, instead of triggering player hotkeys.
W3C documentation for tab keyboard navigation: https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-automatic/


masaball (approved):
:+1: Looks good

PR #448, opened 2024-03-11
Fix Video.js custom components crashing when initializing

Dananji:
Fix for Video.js custom components crashing sometimes when initializing the Video.js instance with Canvas switching.


masaball (approved):
:+1:

Issue #401, opened 2024-02-09
Upgrade VideoJS to 8.0 or latest stable version


Dananji:
### Description

Video.js version in Ramp has been pinned to 7.* because it needed Video.js to be that version to use the quality selector plugin used in Ramp.
Ramp was waiting on https://github.com/silvermine/videojs-quality-selector/pull/93 to be merged and released to upgrade Video.js. Now that it has been merged and released `@silvermine/videojs-quality-selector` can be upgraded from  1.2.4 to 1.3.1 which is the latest version of the plugin. And subsequently Video.js can be upgraded from 7.* to 8.*

### Done Looks Like
- [x] `@silvermine/videojs-quality-selector` upgraded to 1.3.1
- [x] `videoj.js` upgraded to latest stable 8.* version
- [x] VideoJSFileDownload component's code changes to use `extends` instead of `videojs.extend(..`
- [ ] Test all the custom VideoJS components (Progressbar, previous/next canvas buttons, current time display in progress bar, zoom scrubber (magnifying glass), file download) works as expected in all supported browsers and devices
- [ ] Test all the other customization Ramp has implemented on top of native VideoJS functionalities work as expected in all supported browsers and devices: Chrome on Android, Safari on iPhone and iPad, Chrome and Safari on desktop

joncameron:
This can be worked on after 7.7 release; will require a retest across media player test cases since this will be a large change.

joncameron:
Most of the work here will be testing; custom components should all be tested for functionality.

Dananji:
This work has been deployed to Ramp demo site

joncameron:
Need to update test cases in Testomat.io to make sure all of the areas mentioned in this ticket are represented.

elynema:
@charumitraravi This should be able to be tested on avalon-dev. The QA specifically covers the custom components that have been added to Ramp (file download is only available on the Ramp demo site). But it would be good to note that all functionality in the player is working. This is good preparation as we move into regression to QA this well, and I don't think it has been tested yet.

charumitraravi:
Tested all the facets in the media player (forward, backward, pause, scrubber movements, captions, magnifying glass,volume,quality,speed and download in ramp site)and the hotkeys, across the web (chrome,safari) , iphone , ipad (safari) and android(chrome) in different orientations and in full screen . 

- Caption is switched on by default. If I switch off the captions, the selection is not remembered when the player advances to the next section of the item. 
[...]

PR #453, opened 2024-03-15
Upgrade Video.js to v8.10.0

masaball:
Merge after 3.1.0 release

Related issue: #401 

This PR upgrades Video.js to 8.10.0 and @silvermine/videojs-quality-selector to 1.3.1. 

In my testing, all of the custom components seemed to be working as I would expect on all platforms once the File Downloader component had been updated to the new format required by Video.js. 

I did encounter a change in behavior with captions on mobile devices where the text was covering the video. In Video.js 8.4.0 they implemented a [change to force text tracks to overlay the video player](https://github.com/videojs/video.js/pull/8009). Our custom CSS to override text track size resulted in the overlaid text track being too large in font so I have removed that CSS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,9 @@ yarn add @samvera/ramp
 yarn add video.js@7.21.3
 ```
 
-**NOTE**: `video.js@7.21.3` needs to be used until the [fix](https://github.com/silvermine/videojs-quality-selector/pull/93) to use the latest Video.js (v8.0.4) in `@silvermine/videojs-quality-selector` is merged and released.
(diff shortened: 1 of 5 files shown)


Dananji (approved):
Looks good as far as I can tell 👍 

One thing I noticed missing, is updating the  installation instructions in`README.md`, where we have videojs version pinned to 7.*. 

I'm not sure whether we should update this now or later when we release this change. What do you think?

masaball:
I'm not sure either. It is probably fine to leave the note alone for now and update it at the next release, but then we have to remember to update it. If we want to update now, maybe we could change the note to something like `NOTE: Ramp versions <= 3.1.3 contain libraries and code that use functions deprecated in Video.js >= 8.0. Installers must use video.js@7.21.3. `

Dananji (approved):
👍

Issue #452, opened 2024-03-14
[BUG] All captions are highlighted as selected


Dananji:
### Description

In the captions menu in Ramp, all captions are shown as selected even though only one caption is selected.
![Screenshot 2024-03-14 at 4 43 50 PM](https://github.com/samvera-labs/ramp/assets/1331659/3361d8b4-24e8-4477-ad6c-ca2688e38613)

### Environment
- All desktop browsers (Safari, Firefox, Chrome)
- Chrome and Safari in iPad
- Chrome in Android

### Done Looks Like

- [ ] Only the selected caption is shown as selected on player initialization

### Notes
Ramp only sets `default` attribute to `true` for the first caption in the list, but Video.js makes 2 passes over the text tracks list and marks multiple captions as selected in the captions menu when initializing.

elynema:
Test on demo site before we make a new Ramp PR.

charumitraravi:
Verified this item on the Ramp demo site on Chrome and Safari browsers, on web and mobile. The fix is working as expected.

PR #455, opened 2024-03-18
Fix multiple captions being selected on initial load

Dananji:
Related issue: #452 

On initial load programmatically select the first available caption only one caption is shown as selected in the captions menu;
Screenshot 2024-03-18 at 12 36 59 PM

Additional CSS fix to captions icon: fixed captions menu shifting down when captions are turned off;

Before:
Screenshot 2024-03-18 at 12 41 02 PM

After:
Screenshot 2024-03-18 at 12 41 26 PM


masaball (approved):
:+1:

PR #459, opened 2024-03-22
Prevent structure items getting highlighted for inaccessible items

Dananji:
This PR fix a regression from the work done in https://github.com/samvera-labs/ramp/pull/447.
Regression: for inaccessible items all structure items are getting highlighted since there are not active items. Because of this, the check for setting active items in the styling was inaccurately evaluating the condition to be true and highlighting all the structure items as active when none of them are.

Before: 
Screenshot 2024-03-22 at 11 18 06 AM

After:
Screenshot 2024-03-22 at 11 17 37 AM


masaball (approved):
Good catch!

Issue #458, opened 2024-03-20
Spans Past Duration Do Not Display As Links in Structural Metadata


joncameron:
# Description
Existing structure is not displaying correctly in the structural navigation component on the item page when spans go beyond the duration of the file. Spans are active like divs and not displaying as hyperlinks that navigate to that a time point in the media player.

Example: https://purl.dlib.indiana.edu/iudl/media/b68x31xp9v

## Previous Investigation (see https://github.iu.edu/AvalonIU/Avalon/issues/233)
Further looking around and it looks like the integer durations were a red herring/coincidence. [This item](https://media.dlib.indiana.edu/media_objects/c534fq39c) has a broken final structure link because total duration is 2912.314 and end of last segment is 2912.32. So the problem is just structure items having ends past the end of the object duration. Duration assignment and serialization is likely fine.

# To Reproduce
- Go to ramp.avalonmediasystem.org
- Use https://avalon-dev.dlib.indiana.edu/media_objects/fn106x94r/manifest.json as the manifest source
- Track 9 in the 2nd section does not display as a link

# Done Looks Like
- [ ] Spans with end times that go beyond the duration of the stream are active as links, and selecting them seeks the player to the beginning of the span
- [ ] Spans with begin times and end times past the duration of the stream appears as links but are a no-op (no action upon selection)

# Screenshots

![Screenshot 2024-03-18 at 1 43 38 PM](https://media.github.iu.edu/user/2401/files/5f4ef2fc-73b8-4b8f-8f1c-d559d81d13d8)
 
![Screenshot 2024-03-18 at 1 44 11 PM](https://media.github.iu.edu/user/2401/files/221c54a5-89a7-4311-9e9a-583a147cb31e)

joncameron:
Will need to thoroughly test the structure as it shows up in the player (highlighting) to make sure it's working correctly.

elynema:
This is unlikely to affect other Avalon institutions, as most don't use structural metadata as heavily as we do at IU, so this could be pushed to next full release and cherrypicked out to MCO.

Dananji:
Testing notes for done looks like criteria;

> Spans with end times that go beyond the duration of the stream are active as links, and selecting them seeks the player to the beginning of the span

Test item: Last timespan in the second section in https://avalon-dev.dlib.indiana.edu/media_objects/fn106x94r/manifest.json

>  Spans with begin times and end times past the duration of the stream appears as links but are a no-op (no action upon selection)

[...]

PR #471, opened 2024-04-04
Adjust marker display on UI when end > canvas duration

Dananji:
Before:
Screenshot 2024-04-03 at 10 05 29 AM

After:
Screenshot 2024-04-03 at 10 05 04 AM

This doesn't change the behavior but the appearance of the marker on the time rail for structure timespans that exceed the Canvas duration.
Related issue: #458


masaball (approved):
:+1:

Issue #467, opened 2024-04-02
SRT Files Download as VTT


joncameron:
# Description
Avalon's /captions route is automatically transforming SRT files to VTT for display in the client. The download from Ramp is in turn returning a VTT instead of the original SRT when a user downloads from the Transcript component. When downloading from the Ramp Transcript component, the original SRT file should be the one downloaded to the client.

# Implementation Strategy
- Captions are added as two annotations: supplementing (display in player, currently used), and rendering (to be used for download) - https://github.com/avalonmediasystem/avalon/issues/5725
  - Files component only shows rendering annotation
  - Supplemental file URL serialized into the manifest as an rendering annotation

That what we would show in Ramp would be set as 'rendering' would work with the ethos of IIIF.

# Done Looks Like
- [ ] Ramp only displays the supplementing caption in the player
- [ ] Ramp only displays the rendering caption in the Files/Supplemental Files component
- [ ] Supplementing annotations do not display in the Files/Supplemental Files component
- [ ] Update Ramp documentation - this will change existing behavior for the Supplemental Files component

# Notes
- Route for captions: supplemental_files/sdfli2398/captions # referenced in manifest
- Captions endpoint is only returning VTT
- supplemental_files/sdfli2398 should return the original SRT

Dananji:
This can be tested on Ramp demo site.

charumitraravi:
I tested this item and I am able to download srt captions in the original format. However, these are the following issues I encountered with srt transcripts:

1. When I try to access an item that only has an SRT transcript (for example, [this page](https://avalon-dev.dlib.indiana.edu/media_objects/4x51hj01t)), the webpage crashes.

2. If a media item has multiple transcripts available and one of them is an SRT file, the page also crashes when I select the SRT transcript from the list (as seen in [this example](https://avalon-dev.dlib.indiana.edu/media_objects/g158bh28p)).
(refer the screenshots attached)

<img width="1086" alt="Screenshot 2024-05-16 at 7 45 08 PM" src="https://github.com/samvera-labs/ramp/assets/52535309/ca5d23f9-5f33-475c-ba91-2bfb22b797d3">
<img width="856" alt="Screenshot 2024-05-16 at 7 47 59 PM" src="https://github.com/samvera-labs/ramp/assets/52535309/6a94a648-3009-440c-9e22-f05478a6bf8e">

joncameron:
SRT bug issue: https://github.com/samvera-labs/ramp/issues/489

PR #474, opened 2024-04-04
Only display rendering files in the supplemental files component

Dananji:
> **Need` to test with a sample manifest from Avalon once the manifest generation work is deployed to `avalon-dev`** 

✅ 

In the Ramp demo site sample;

Before:
Screenshot 2024-04-04 at 3 01 42 PM

After:
Screenshot 2024-04-04 at 3 02 04 PM

Related issue: #467


masaball (approved):
Looks good as far as I can tell

PR #478, opened 2024-04-10
Build active structure item classname into a single line

Dananji:
This PR fixes a bug introduced in https://github.com/samvera-labs/ramp/pull/447.

This bug was released in version 3.1.0, and possibly breaks add to playlist functionality for current track option in Avalon 7.7.1.


cjcolvar (approved):
Good catch!

elynema:
Tested add to playlist functionality on demo and avalon-staging. The "current track" option was disabled on demo site (7.7.1), but worked fine on avalon-staging in Chrome and Safari.

PR #481, opened 2024-04-19
Skip combination keypresses in player hotkeys

Dananji:
Skip combination key presses with `Ctrl` / `Cmd` / `Alt` / `Shift` keys when invoking player hotkeys. e.g. `Ctrl` + `F` for search on browser page.


masaball (approved):
:+1:

Issue #399, opened 2024-02-09
[BUG] Full screen playback is not retained when auto-play to next canvas


elynema:
# Description

As of 7.6 in Avalon if a user started playback on a section on the media item page and entered full-screen mode, the full-screen would be retained when advancing to the next section.

This is no longer happening. Instead, at the end of the section, the browser automatically exits full-screen and playback on the next section begins in the browser page, not in full-screen mode.

# To Reproduce

1. Start playback of a multi-section video item
2. Enter full-screen mode
3. Allow the video to automatically advance to the next section

It is expected behavior that the browser should remain in full screen mode.

# For QA

 - [x] test transitions audio to video and video to audio
 - [x] check that all custom components load properly with canvas change
 - [x] test transitioning to and from inaccessible items in playlist / media object page
 - [x] check sticky quality selector and volume when transitioning between canvases
 - [x] test cookbook example with multiple files in one canvas
 - [x] test CC value changes between canvases
 - [x] Test values in Transcript / Files tabs when switching canvases
 - [x] test that zoom scrubber works after transitioning canvases / playlist items
 - [x] test these transitions with prev/next canvas button as well as selecting section / timespan in structured navigation
 - [x] Test markers and source item display changes when switching playlist items
 - [x] test that active segment / section markers transition when changing canvas / playlist item
 - [x] Test transition when encountering an inaccessible item in full screen
 - [x] Caption button shows in player both in playlist context and in media item (even if the first file doesn't have captions and the second file does) 
 
# Question
 - [ ] when in full screen mode, and playlist reaches an inaccessible item, what should happen? What used to happen?
 Reasonable first pass behavior is for full screen to exit, go back to playlist/media object page, and then advance to next item not in full screen after 10 seconds.

Ideally, we could retain full screen with the error message showing in full screen and a countdown timer.

joncameron:
This could be a big change because we rely on a lot of communication from Video.js. For this to be implemented we need to retain the Video.js instance; this could break other parts in the process.

elynema:
[...]

PR #484, opened 2024-04-26
Fix track scrubber bugs in Videojs re-setup work

Dananji:
Related issue: #399


cjcolvar (approved):

PR #488, opened 2024-05-01
Don't include 2nd bundled copy of css on demo page. Resolves #487

### LINKED ISSUE #487: Duplicate css being served on demo page breaks css
**Describe the bug**
Currently there is duplicate CSS being served on the `/demo` page.
In `./demo/app.js` there is an import to `../dist/ramp.css`. That's bundled copy of all the CSS. But it's also out of date and needs to be manually built. If you make a change to existing CSS, it will have no effect because the bundled CSS is further down on the page.

**To Reproduce**
1. Run `yarn demo` and open the demo site http://localhost:3003/ 
   image
  here you can see that the download transcript button is currently green
2. Open `./src/styles/_vars.scss`
3. Change 

    ```scss
    $primaryGreenDark: #2A5459;
    ```
    to
    ```scss
    $primaryGreenDark: red;
    ```
    the hit save
4. Back in your browser, the button remains green, even after hard refresh.
 
**Expected behavior**
The button turns red. 

**Screenshots**
code

patrick-lienau:


Dananji (approved):
Looks great. Thank you for the PR 🎉 

Just one suggestion. It seems there are extra tab/space characters in the code. Would you please be able to re-format it?

patrick-lienau:
No problem. It's now formatted with 2-space indents.

PR #491, opened 2024-05-06
Set player container aspect ratio to 16:9 for blank canvas messages

Dananji:
Always set player container's aspect-ratio to video, when displaying a blank Canvas (inaccessible item).


cjcolvar (approved):
Nice catch!

PR #493, opened 2024-05-07
Accessibility fixes to issues found in axe report

Dananji:
Found these issues via a axe scan on Ramp demo site;
- Ensures `<dt>` and `<dd>` elements are contained by a `<dl>`. Description list item does not have a `<dl>` parent element in `MetadataDisplay` component
- Ensures elements with an ARIA role that require child roles contain them. Element has children which are not allowed: li[aria-label] in the `StructuredNavigation` component
- Ensures every HTML document has a lang attribute. The <html> element does not have a lang attribute in the demo implementation page

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -1,4 +1,4 @@
-<html>
+<html lang="en">
 
 <head>
   <title>Ramp</title>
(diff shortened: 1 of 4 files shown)


cjcolvar (approved):
Looks good to me.

PR #495, opened 2024-05-09
Remove CSS classes for captions, fix CSS for placeholder canvas messages

Dananji:
Follow-up on suggested changes in https://github.com/samvera-labs/ramp/pull/460#discussion_r1576531153, fix CSS for placeholder canvas messages with links, and cleanup unnecessary code left from https://github.com/samvera-labs/ramp/pull/460.


cjcolvar (approved):
Looks good!

Issue #464, opened 2024-03-27
Add search box to Transcript tab


elynema:
# Description

Create a new component in Ramp to provide transcript search. If we decide to extend search to metadata and/or other annotations, we can pull this component out from nesting within transcript component and let it stand on its own.

User can enter / refine a single search term. We do not need to support the ability to add multiple independent search terms like Aviary supports.

Likely, this search box will be wired up to back-end Content Search API, but need to finalize that before beginning on tying this to back-end.

# Done Looks Like

- [ ] Search box is located above the drop-down on trannscript tab (like Alexander Street Press)
- [ ] Test in search box reads "Search transcript"
- [ ] Transcript search should search content of all available transcripts for current canvas.
 - [ ] Below / next to search box, show number of hits with arrows to page through them in transcript text
 - [ ] Drop-down to select transcript shows # of hits in each transcript (like Aviary)
 - [ ] Link to Clear search available after searching

joncameron:
This issue needs a mockup! We have visual examples to draw from in the Google document about transcript display and searching.

joncameron:
Should this be moved to the Ramp repository?

joncameron:
@joncameron to write a separate issue for extending the drop-down listing of transcript documents and changing the download button. Styling also needs to be considered and changed from the mockup to match existing styling and for accessibility.

patrick-lienau:
Last year my company forked ramp to implement similar functionality for a client. We had another project come up and I never got the opportunity to PR my work. Since then, the upstream has changed enough to make a direct PR difficult so in the past week I've gone and reimplemented the functionality off the current codebase. 

![image](https://github.com/samvera-labs/ramp/assets/115490089/aed71a5f-9289-4e2f-9660-86379b1ec6b0)
![image](https://github.com/samvera-labs/ramp/assets/115490089/136df5aa-5add-4df8-9372-77814b3cfae7)
![image](https://github.com/samvera-labs/ramp/assets/115490089/828d474f-cc00-4e20-ad2c-a711fe0318e0)

@joncameron please know that we do intend to contribute this functionality in the very near future. I believe we have a meeting on the 17th regarding this.

elynema:
@patrick-lienau Any chance you can drop us a PR with these front-end code chages before May 17? We'd like to take a look at it.

elynema:
@joncameron @Dananji Took a quick look on the demo site at how the contributed code works, and I see a couple of things to decide how to handle:

[...]

PR #497, opened 2024-05-10
Transcript Search Functionality

patrick-lienau:
As promised in #464, here is a PR for the addition of search functionality to transcripts.

Based on the list of checkboxes in #464, I think we're largely in alignment with what such an implementation should look like in terms of UI/UX. 

##### State before any search term entered
image
##### Previous/Next button, "Result X of Y", and result match highlighting
image
##### Search Results as markers on progress bar
image
##### Mobile View
image

I've tried to limit the scope of the changes necessary to achieve this as much as possible. That being said, in order to allow for highlighting matches in transcript lines (ie: making the search terms bold/underlined) and to enable the prev/next button to scroll the transcript viewer, some significant changes had to be made to `<Transcript />`. However, upon review, I think you'll agree that the changes will result in a more maintainable component.

### Customization

I know there was some interest of connecting this search functionality to some IIIF search API. Since we already have the full text of the transcript in-browser I'm not sure how much benefit that would add in practice.. However, I did take that into account. The `<Transcript />` component can be given a custom "matcherFactory":

```jsx
const CustomTranscript = ({ transcripts }) => { 
  const matcherFactory = useMemo(() => ( // useMemo is for referential stability
    (transcriptItems) => {
      return async (query, abortController) => {
        try {
          const res = await fetch('https://my-api.com/search?q='+query, {
            signal: abortController.signal
          });
          const json = await res.json();
          return res.results;
        } catch (e) {
          console.error(e);
          return [];
        }
      }
    }	
  ));
  return (
    <Transcript
      playerID="player-id"
      transcripts={transcripts}
      search={{
        matcherFactory
      }}
    />
  );
};
```
It supports both async results AND gives an `AbortController` which can be passed to `fetch`. So If there  is a api request in progress with the query `hello wo` and the user types, updating the query to `hello wor`, the previous fetch is canceled by the browser. This should make it possible to wire up to an external API or integrate with some fuzzy search library of your preference. By default, it behaves like your browser's "Find in Page" (<kbd>CTRL</kbd>+<kbd>F</kbd>) search and seems to work quite well.

### Next Steps

I know there are use cases for RAMP that I'm not really aware of. I'm really just going off the demo page so I'm not sure how this all interacts with thing like Playlists, and whatever the Markers tab does. Hopefully once you've had a chance to review, we can discuss where I might be able to find some examples of those so I can test their interaction.

I've written several tests, but have held off on going for more complete coverage, updating the documentation, extensively annotating code, etc until your team has had a chance to take a look at it. Once the implementation has undergone any requested changes, I'll go through, finish writing tests, annotate code, update the docs, etc.

Anyway, I'd love any feedback anyone has. Thanks.


demo/app.js
```diff
@@ -153,6 +153,7 @@ const Tabs = ({ tabValues, manifestUrl }) => {
       <TabPanel id="transcriptsTab" tabId="transcripts" tabIndex={1} activeTab={activeTab}>
         <Transcript
           playerID="iiif-media-player"
+          showSearch={true}
```
Dananji, line 156: As I understood this prop can be used to turn the search functionality on/off.
I couldn't find where this gets read into the code in the transcript component to be used. Could you please point me to the place where this gets used?

src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
```diff
@@ -395,6 +397,60 @@ function VideoJSPlayer({
...
+
+  // pretty sure this can be removed...
+  React.useEffect(() => {
+    const player = playerRef.current;
+    if (playlist.markers?.length > 0 && isReadyRef.current) {
+      // Set player duration, for markers API. The value set in the player update
+      // function sometimes doesn't update the duration in the markers API.
+      player.duration(canvasDurationRef.current);
+    }
+  }, [playerRef.current, isReadyRef.current, playlist.markers]);
```
Dananji, line 410: Yes, you are correct. I think this is redundant.

src/components/MediaPlayer/VideoJS/components/js/VideoJSTrackScrubber.js
```diff
@@ -215,7 +215,7 @@ function TrackScrubberButton({ player, trackScrubberRef, timeToolRef, isPlaylist
       In playlists, markers are timepoint information representing highlighting annotations, 
       therefore omit reading markers information for track scrubber in playlist contexts. 
     */
-    if (player.markers && player.markers.getMarkers()?.length > 0 && !isPlaylist) {
+    if (player.markers && typeof player.makrers !== 'function' && player.markers.getMarkers()?.length > 0 && !isPlaylist) {
```
Dananji, line 218: Good catch!

For avalon manifests, we use player markers to highlight the structure ranges. So, I think we need to refactor this code to filter only the structure range markers, now that we have the transcript search hits using markers as well.
I tested the following code with a manifest from avalon, which seemed to work;
```
if (player.markers && typeof player.markers.getMarkers === 'function' && player.markers.getMarkers()?.length > 0 && !isPlaylist) {
      const track = player.markers.getMarkers().filter(m => m.class == 'ramp--track-marker--fragment')[0];
      if (track?.length > 0 && track[0].key != currentTrack?.key) {
        setCurrentTrack(track[0]);
      }
    }
```

src/components/Transcript/Transcript.js
```diff
@@ -1,33 +1,226 @@
 import React from 'react';
+import cx from 'classnames';
```
Dananji, line 2: Is there a particular reason to use this library instead of using vanilla JS to build conditional classnames? I have not used this library before, so I was curious.

src/components/Transcript/Transcript.js
```diff
@@ -1,33 +1,226 @@
...
+    );
+  } else {
+    return (
+      <p key="no-transcript" id="no-transcript" data-testid="no-transcript" role="note">
+        {transcriptInfo.tError}
+      </p>
+    );
+  }
+};
+
```
Dananji, line 212: This is much cleaner than what we had before code-wise. Looks great 🎉

src/components/Transcript/Transcript.js
```diff
@@ -89,32 +285,21 @@ const Transcript = ({ playerID, manifestUrl, transcripts = [] }) => {
...
-                  // remove highlight
-                  tr.classList.remove('active');
-                }
-              }
-            });
+          setCanvasIndex(cIndex);
+          setCurrentTime(playerRef.current.currentTime);
+
+          playerRef.current.addEventListener('timeupdate', () => {
+            setCurrentTime(playerRef.current.currentTime);
```
Dananji, line 302: This is not a blocker for this PR. This is more of a question for any future work, if we decide to not to make the transcript component an independent component.

Is it possible to remove this interval, which is used to keep track of the player instance on the page since the transcript component is not wrapped within the player context provider? 
The reason for this question is; with the work done in this PR the PlayerContext seems to be wrapping (at lease partially?) the transcript component, so we can use the player context to get this information instead of this interval? Am I understanding this correctly?

src/components/Transcript/Transcript.js
patrick-lienau: `classnames ` is purely for ergonomics. It helps avoid long ternaries or akward interpolation when you have multiple conditional classnames. It [weighs virtually nothing](https://bundlephobia.com/package/classnames@2.5.1) but can be removed for vanilla JS, I probably just included it out of habit.

Dananji (commented):
This looks really good. Thanks for contributing your work upstream.

I just have a couple of questions about the code.

src/components/Transcript/Transcript.js
Dananji: That sounds good. I think there are multiple places in the code, where we use conditional classnames with long conditions. And I think this could be a nice addition to the code, which will improve the readability and maintenance of the code going forward 👍

src/components/Transcript/Transcript.js
patrick-lienau: IIRC, the `currentTime` property of `playerContext` is only updated when you use structured navigation (and possibly some other places) and does not update during playback.  You could certainly update that value during playback to make it the "source of truth" so that it is in sync with the player by adding something like this presumably to `VideoJSPlayer.js`:

```jsx
useEffect(() => {
  if (!playerRef.current) return;
  playerRef.current.addEventListener('timeupdate', () => {
    playerDispatch({ type: 'setCurrentTime' , currentTime: playerRef.current.currentTime })
  });
}, [playerRef.current]);
```

However that would have the undesirable effect of triggering a full re-render your entire tree several times a second even when the user is doing nothing. Since you're using react reducers and not redux you can't rely on selector memoization either... I'm not sure if hoisting the currentTime to the root is the best approach but even if you did, honestly, I don't think the performance would be too bad if it was denounced enough. A resolution of a quarter second (250ms debounce) should should be aggressive enough but I haven't done any testing to say for sure.

patrick-lienau: IIRC, the `currentTime` property of `playerContext` is only updated when you use structured navigation (and possibly some other places) and does not update during playback.  You could certainly update that value during playback to make it the "source of truth" so that it is in sync with the player by adding something like this presumably to `VideoJSPlayer.js`:

```jsx
useEffect(() => {
  if (!playerRef.current) return;
  playerRef.current.addEventListener('timeupdate', () => {
    playerDispatch({ type: 'setCurrentTime' , currentTime: playerRef.current.currentTime })
  });
}, [playerRef.current]);
```

However that would have the undesirable effect of triggering a full re-render your entire tree several times a second even when the user is doing nothing. Since you're using react reducers and not redux you can't rely on selector memoization either... I'm not sure if hoisting the currentTime to the root is the best approach but even if you did, honestly, I don't think the performance would be too bad if it was denounced enough. A resolution of a quarter second (250ms debounce) should should be aggressive enough but I haven't done any testing to say for sure.

demo/app.js
patrick-lienau: Ah, I forgot to update that line actually. Originally I had the various search props spread out on the `Transcript`
component like so: 

```jsx
<Transcript searchEnabled={true} matcherFactory={myCustomMatcher} />
```

Just before I did the PR did some cleanup. Now all the search things go inside a single prop: `search`
```jsx
<Transcript search={false} />
```
```jsx
<Transcript search={{ matchesOnly: true, matcherFactory: myCustomMatcher }} />
```

src/components/Transcript/Transcript.js
Dananji: Yes, that makes sense. Thanks for clarifying it. Apologies for the confusion.

In that case, I think the last change you made is not needed for this work. We can keep the Transcript component wrapped within the `PlayerContext` for this PR. 
And if we need to keep the Transcript component independent of the central state (need to discuss with the team), we can make the change as needed in the future.

So once you undo the last commit, I think we are ready to merge this PR 🎉

demo/app.js
Dananji: Could you please update the code to reflect the changes? Thanks!

src/components/Transcript/Transcript.js
patrick-lienau: If I revert that commit, the styleguide page will be broken because it isn't wrapped in context. The way I did the change in c7f9c08, it will check if it has dispatch via context, and if not, then it won't move the playhead when a line is clicked/won't markers to the progress bar but will otherwise work. So I can revert and make it have a hard fail if it's not wrapped but I don't see the benefit of that. 

that being said, I'll revert that commit momentarily

demo/app.js
patrick-lienau: [1b144a4](https://github.com/samvera-labs/ramp/pull/497/commits/1b144a413cd5de34911be49a72c12aa3f53a667a) cleans that up

src/components/Transcript/Transcript.js
Dananji: Thanks!

Yes, makes sense to just do that at this stage. We could make changes as necessary in the future.

I checked the styleguide docs and they seem to be working with the latest code 👍

Dananji:
Thanks for making the changes and answering my questions.

I have one more question on the functionality of the search markers in the progress bar; 
When you click on a search hit marker on the progress bar it doesn't scroll the transcript component to that respective transcript line, it feels like this is broken, am I right?

patrick-lienau:
That's some functionality I had thought about but never implemented. It would require being able to pass a `selectedMarkerId` (or something like that) from the `<VideoJSProgress />` component into to the `<TranscriptList />` component.

Markers were a request from our client but I wasn't even sure how your team would feel about the markers because I don't believe they were mentioned when we spoke before. I didn't want to build that out when it could have been requested that they be removed from the PR.

Perhaps this could be a future addition?

Dananji (approved):
Looks great! 👍 

It'd be great if you could squash and merge to keep Git history clean and concise. Thanks!

Dananji:
Related issues: https://github.com/avalonmediasystem/avalon/issues/5753, #464

patrick-lienau:
> and merge to keep Git history clean and concise. Thanks!

image


Alas, you'll need to handle the merge itself  ↑↑

Dananji:
@patrick-lienau I sent an invitation adding you as a collaborator with write permissions. Now you should be able to merge 🎉

Issue #494, opened 2024-05-07
Scrubbing slow in Safari in MCO, AVAnnotate


elynema:
# Description

Scrubbing seems to be responding quite slowly particularly in Safari (tested on desktop). I notice this particularly when dragging the scrubber bar; it seems to be faster to just point at a spot in the progress bar and click. In MCO and in Ramp, when dragging the scrubber in Safari, it seems to take 4-7 seconds for the video to re-start, which is a lot. This behavior was noticed by Jon Dunn when he was working on bringing MCO content into AVAnnotate, which is using the Aviary player. When testing in Safari he noticed this slowness in both MCO and AVAnnotate. Note that Aviary seems to use video.js, so that could possibly explain similar behavior across multiple tools. 

In addition, in Ramp in Safari, when scrubbing, the player will spin, jump to a frame, spin, and possibly repeat those steps once or twice before actually starting playback again.

Example record in MCO: https://media.dlib.indiana.edu/media_objects/3r075k73g
Record in Ramp: https://ramp.avalonmediasystem.org/?iiif-content=https://media.dlib.indiana.edu/media_objects/3r075k73g/manifest.json
Record in AVAnnotate (using Aviary player): https://elynema.github.io/avalon-77-test/afrique-sur-seine
Record in Aviary IIIF Player: https://iiif.aviaryplatform.com/player?manifest=https%3A%2F%2Fmedia.dlib.indiana.edu%2Fmedia_objects%2F3r075k73g%2Fmanifest.json&tab=Annotations

# Done Looks Like

 - [ ] test scrubbing with these different tools in Safari and Chrome and write down what is found
 - [ ] if we can identify a persistent slowness in Safari, investigate errors in console to see if can identify what the issue might be
 - [x] Goal would be 1-3 seconds response time when scrubbing (max)
 
# QA
 - [ ] test scrubbing in iOS native player, as well as Safari on desktop

joncameron:
Could be related to Video.js; may be due to HLS + Video.js.

Dananji:
I tested the same MCO item in the following viewers in Safari;
- Ramp (uses Video.js): 
  - On MCO: 
     - Click on a timepoint -> slight time lag when updating the frame 
     - Dragging the scrubber -> updated the frame right away but then it jumps back to the previous timepoint (where the scrubber was when dragging started) and dragging is stopped and jumps to the new timepoint (where the dragging ended)
[...]

PR #504, opened 2024-05-22
Fix seeking in safari

Dananji:
Related issue: #494

Changes in the PR:
- Block player until `loadedmetadata` event is fired, so that the user cannot start scrubbing which can lead to player freeze
- Use a cancellable interval for time updates in Safari browsers

### DIFF
--- a/src/components/MediaPlayer/VideoJS/components/js/VideoJSProgress.js
+++ b/src/components/MediaPlayer/VideoJS/components/js/VideoJSProgress.js
@@ -3,7 +3,7 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import videojs from 'video.js';
 import '../styles/VideoJSProgress.scss';
-import { IS_MOBILE, IS_IPAD } from '@Services/browser';
+import { IS_MOBILE, IS_IPAD, IS_SAFARI } from '@Services/browser';
 
 const vjsComponent = videojs.getComponent('Component');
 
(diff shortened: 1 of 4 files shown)


src/components/MediaPlayer/VideoJS/components/js/VideoJSProgress.js
```diff
@@ -240,6 +240,19 @@ function ProgressBar({
...
+      let pressedKey = event.which;
+      if (pressedKey === 37) {
+        event.preventDefault();
+        setProgress(player.currentTime() - 5);
+      }
+      if (pressedKey === 39) {
+        event.preventDefault();
+        setProgress(player.currentTime() + 5);
+      }
+    });
```
masaball, line 255: How does this interact with the left/right arrow handling in the playerHotKeys function? Depending on what order the functions fire in, this seems like it could potentially cause the `progressRef` to be offset by 5 seconds.

src/components/MediaPlayer/VideoJS/components/js/VideoJSProgress.js
```diff
@@ -286,33 +304,72 @@ function ProgressBar({
...
+    timeToolRef.current.innerHTML = formatTooltipTime(currentTime);
 
     handleTimeUpdate(initTimeRef.current);
   }, [player.src(), targets]);
 
+  /**
+   * A wrapper function around the time update interval, to cancel
+   * intermediate updates via the time interval when player is 
+   * waiting to fetch stream
+   * @param {Number} delay time interval in milliseconds
```
cjcolvar, line 316: Should this be removed? I don't see the param mentioned in the method.

src/components/MediaPlayer/VideoJS/components/js/VideoJSProgress.js
Dananji: `progressRef` variable is used only to update the progress bar UI. While any updates to `player.currentTime()` update `progressRef` variable it doesn't happen the other way around. So it wouldn't offset the time by 5 seconds. 

However, this is only relevant to Safari because the `timeUpdateHandler` function skips this update for Safari.
Maybe this eventListener can be scoped further? Let me take look at that again.

Dananji: You were right, it does offset the time. Thank you for catching it 🎉 
As it turns out we don't need that eventListener at all!

PR #505, opened 2024-05-23
Enable/disable captions icon as needed when switching between canvases

Dananji:
Related issue: #399 

Fixes the bug reported in https://github.com/samvera-labs/ramp/issues/399#issuecomment-2121198181
> The CC button is not showing in playlists. The captions themselves are displaying, but no CC button in the player to turn them on/off. It's working fine on media items.

This bug is persistent in media object manifests, when the first Canvas in the Manifest doesn't have captions but a latter has captions.


Dananji:
node-14 tests are failing since `replaceAll()` used in transcript search automated tests is supported from [Node 15](https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V15.md#15.0.0) onward.

Rebase this after #506 is merged.

src/services/iiif-parser.js
```diff
@@ -706,9 +706,11 @@ export function getStructureRanges(manifest, isPlaylist = false) {
       } else {
         hasRoot = true;
         // Total duration for all resources in the Manifest
-        manifestDuration = canvasesInfo.reduce(
-          (duration, canvas) => duration + canvas.range.end, 0
-        );
+        if (!isPlaylist) {
```
cjcolvar, line 709: Why is this check necessary now?  It looks like the conditional on line 693 should keep any playlists from reaching this code.

cjcolvar (approved):
Looks good to me.  I had one question but I think it doesn't need to block merging.

src/services/iiif-parser.js
Dananji: Yes, I'm not sure why I put it there. Thanks for catching it.

PR #506, opened 2024-05-23
Update Node in Circle-CI

Dananji:
Found that both node version in Ramp's CircleCI orb are EOL, when came across the failing tests in #505. 

So, this PR is to uppgrade Node in Circle-CI from node 14 and 16 to 18 (EOL: April 2025 according to https://endoflife.date/nodejs) and 20.

I first tried bumping up to Node 21, which failed because of [`node-sass`](https://www.npmjs.com/package/node-sass) being deprecated.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,10 +6,10 @@ workflows:
   test-matrix:
     jobs:
       - node/test:
-          name: node-16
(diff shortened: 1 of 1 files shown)


masaball (approved):
:+1:

PR #507, opened 2024-05-23
Fix a couple of minor bugs

Dananji:
Fixes the following bugs;
1. Player marker filtering bug in track scrubber with the addition of transcript search hit markers on the progress bar. The player crashes when switching from a Canvas without structure to a Canvas without structure, because `currentTrackRef` variable in track scrubber component is not getting setup properly. This is a regression of the work done in #497.
2. Avoid unnecessary `console.error` message in the console when using `startCanvasId` prop without specifying a value for `startCanvasTime` prop when Ramp is initialized.


Dananji:
Need to rebase after #506 is merged.

cjcolvar (approved):
Looks good!  I like the how you were able to DRY things up.

PR #518, opened 2024-06-11
Improve VTT parsing

### LINKED ISSUE #499: Transcript component parses 'note' within cue-text as plain text
### Description

Transcript component parses the `NOTE` block in a WebVTT file as plain text and displays in the component. The parse is set to identify these blocks in the header as well as in between the cues in the file.
When there's `note` text within the cue's text with a line break in front of it, the parser recognizes it as a `NOTE` block in the WebVTT file instead of as part of the cue.

### To Reproduce
1. Go to https://avalon-dev.dlib.indiana.edu/media_objects/gf06g267d/
2. Click on second section named 'The Legislative Organization of Congress'
3. Select 'Transcripts` tab
4. Scroll down to transcript cue with the timestamp `00:24:21`
5. See the text for that cue is broken into 2 lines, where the text 'note that neither the Majority' is not a clickable transcript cue, but a plain text display

### Expected behavior
Whole cue text:
- [ ] needs to be contained within a single `<span>` element
- [ ] needs to be linked (clickable) and highlighted when user interacts with the cue
- [ ] is highlighted when playback reaches the respective timestamp

### Screenshots
Example cue;
```
00:24:21.400 --> 00:24:24.370 align:middle line:84%
It's interesting to me to
note that neither the Majority
```
This is displayed in the Transcript component as follows;
Screenshot 2024-05-14 at 1 22 07 PM

### Additional context
Related previous work: #440

cjcolvar:
Fixes #499 and improves handling of NOTE comments.  Also includes spaces that were missing between lines of a cue or comment.

I removed `cleanTimedText` because I needed the new lines to flow through to `groupTimedTextLines` where I added additional handling of NOTE comments.

### DIFF
--- a/src/services/transcript-parser.js
+++ b/src/services/transcript-parser.js
@@ -562,8 +562,7 @@ export function parseTimedText(fileData, isSRT = false) {
     noteLines = notes;
   }
 
-  const timedText = cleanTimedText(cueLines);
-  const groups = groupTimedTextLines(timedText);
+  const groups = groupTimedTextLines(cueLines);
   // Add back the NOTE(s) in the header block
   groups.unshift(...noteLines);
   groups.map((t) => {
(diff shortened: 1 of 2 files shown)


src/services/transcript-parser.js
```diff
@@ -693,22 +677,24 @@ function cleanTimedText(lines) {
...
-    if (line.includes('-->') || (/^NOTE/).test(line.toUpperCase())) {
-      const isNote = (/^NOTE/).test(line.toUpperCase());
+    if (line.includes('-->') || (/^NOTE/).test(line)) {
+      const isNote = (/^NOTE/).test(line);
       t.times = isNote ? "" : line;
       t.tag = isNote ? TRANSCRIPT_CUE_TYPES.note : TRANSCRIPT_CUE_TYPES.timedCue;
-      t.line = isNote ? line : '';
+      // Make sure there is a single space separating NOTE from the comment for single or multi-line comments
+      t.line = isNote ? line.replace(/^NOTE\s*/,'NOTE ') : '';
       i++;
```
masaball, line 689: Does this still need to be in the loop with adding `i++` to the for loop declaration?

cjcolvar: Yes, it advances to the next line so the while loop test starts with the line after the start of the cue/note.

masaball (approved):

PR #520, opened 2024-06-11
showNotes Transcript prop for displaying NOTE comments (default: false)

cjcolvar:
Resolves https://github.com/avalonmediasystem/avalon/issues/5858


masaball (approved):
Looks good

PR #521, opened 2024-06-12
Render time offsets for cues that begin exactly at 0

### LINKED ISSUE #500: First VTT Cue Timestamp Displayed Improperly
## Example
https://avalon-dev.dlib.indiana.edu/media_objects/fj236208t

cjcolvar:
In Javascript 0 evaluates to false so a begin time of 0 skips rendering the time span by accident.  I switched the guard to check that the typeof the begin time is a number.

```
0 && true // => 0
20 && true // => true
typeof 0 // => 'number'
typeof 0 === 'number' && true // => true
```

Resolves #500

### DIFF
--- a/src/components/Transcript/Transcript.js
+++ b/src/components/Transcript/Transcript.js
@@ -114,7 +114,7 @@ const TranscriptLine = ({
           isFocused && 'focused'
         )}
       >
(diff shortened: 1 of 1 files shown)


masaball (approved):
:+1:

Issue #525, opened 2024-06-13
Can't access full-screen button in Android Samsung devices


elynema:
**Describe the bug**
On Android Samsung devices, the full-screen button is missing from the right corner of the media player. (BrowserStack won't allow me to rotate this device for some reason, so I can't confirm if it is actually missing or just doesn't have enough space).

**To Reproduce**
1. Go to Android Samsung Galaxy device (v14 OS) in Browserstack
2. Open a media object in chrome. Example: https://avalon-dev.dlib.indiana.edu/media_objects/g158bh28p
3. There is no full-screen button in the bottom right of the media player.

**Expected behavior**
The full-screen button should show in the far right of the media player on Android Samsung devices.

**Screenshots**
<img width="369" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN1ZpQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--04b069f47fffec59b6b38bfbaf627078a8ee5d07/screenshot-Samsung%20Galaxy%20S24-14.0.png" alt="screenshot-Samsung Galaxy S24-14.0.png" />

**Smartphone (please complete the following information):**
 - Device: Samsun Galaxy S24
 - OS: 14
 - Browser Chrome

**Additional context**
This problem is not happening in Google Pixel devices.

elynema:
This looks ok on the demo site, but needs to be tested in the avalon context to finish QA.

elynema:
@charumitraravi Can you QA this one?

charumitraravi:
If I increase the display size the full screen button still gets hidden (This is on samsung s23 ultra, which is more wider than most devices.)
(Image 1 - increased by 2 units from default , image 2 - increased 1 unit from default size)
Although , this could just be an edge case , considering users who prefer bigger display of elements / smaller devices


![samsung2](https://github.com/samvera-labs/ramp/assets/52535309/3bec080c-aa52-4177-a18e-c0a902a7990d)


![samsumg1](https://github.com/samvera-labs/ramp/assets/52535309/21fc0677-e2d5-4eb8-82f6-886964948c82)

elynema:
@Dananji Do we have any options for resolving this? I know we are supposed to support up to 200% zoom.

Dananji:
[...]

PR #527, opened 2024-06-17
Mobile icons

Dananji:
Related issue: #525 

The adjusted icon width and font-size in Ramp player in Avalon:
bs_realdroid_Mobile_Samsung Galaxy S23-13 0-1080x2340(1)


cjcolvar (approved):
Looks good!

It looks like the changes in your first commit were reverted in your second commit.  I think it would be nice to squash merge this PR to get rid of the first commit so we don't see it and get confused later.

Issue #501, opened 2024-05-16
Show Player Controls and Duration Remaining for Inaccessible Canvas while Fullscreen


joncameron:
## Description
When in fullscreen on an item or playlist with multiple canvases, switching to an inaccessible item will retain fullscreen and show the standard "you don't have permission to play" text for a duration, after which the next canvas is loaded. However, there is no indicator that the player will automatically move on after a period of time, and the media player controls do not display. 

Ramp should retain player controls for inaccessible item display, and a timer is displayed to the user to indicate that autoplay will continue after a set duration. We already have the timer set as part of the inaccessible item handling in the media player, and that value could be presented in the UI.

## Done Looks Like
- [ ] Ramp retains player control visibility and behavior when showing an inaccessible item (revert Video.js changes)
- [ ] Timer displays as duration and progress shows timer state (i.e. as if a 10 second long item is playing)
- [ ] Auto-play resumes in Safari after playback moves on to the next item

joncameron:
Would be better/easier to move straight to the enhancement version of this, where player controls are retained for inaccessible items and a timer is displayed to the user to indicate that autoplay will continue after a set amount. 

We already have the timer set, and that value could be presented in the UI. This could be as simple as reverting Video.js changes to make sure player controls aren't hidden during this. 

@joncamer to create a new issue or change this one to reflect the change in scope. This new issue would be 2 points.

elynema:
Note that in Safari, when in full-screen playback, after the 10 seconds of an inaccessible item, it attempts to load the next item but auto-playback does not start. This works fine in Chrome.

Dananji:
I was not able to reproduce the full-screen auto advance issue on Safari on both Ramp demo site and in avalon-dev for [this playlist](https://avalon-dev.dlib.indiana.edu/playlists/45). Was this happening in a specific playlist?

Dananji:
I was not able to use player controls to show the 10 second duration as described in the done looks like; the reason for this is when there's an inaccessible item as the first item in the manifest there's no Video.js instance. To create a Video.js instance at least one source URL is needed to be provided for the HTML5 video tag. And for inaccessible items sources list is empty.
Instead I opted to show the 10 second timer like it is shown in Youtube;
![auto-advance-preview](https://github.com/samvera-labs/ramp/assets/1331659/99b65bbc-12fc-4f2f-9c04-cac05953af99)

Dananji:
Notes on the current implementation;
[...]

PR #529, opened 2024-06-20
Display previous/next buttons and timer for inaccessible canvases

Dananji:
Related issue: #501 

It was not possible to implement what was described in the done looks like in the ticket, as Video.js requires at least one source to build the player and its controls. And for inaccessible items the sources list is empty in the Canvas. 

Therefore this UI was implemented, by wiring the same functions the previous/next buttons use in the player controls to the buttons in the display.

When first item is inaccessible only display the next button; (and only displays previous button when last item is inaccessible)
Screenshot 2024-06-24 at 3 38 36 PM

Previous and next buttons when item is in the middle of the list;
Screenshot 2024-06-24 at 3 39 24 PM

Hides the timer display when auto-advance is turned off;
Screenshot 2024-06-24 at 3 38 06 PM

### DIFF
--- a/src/components/MediaPlayer/MediaPlayer.js
+++ b/src/components/MediaPlayer/MediaPlayer.js
@@ -285,6 +285,11 @@ const MediaPlayer = ({
     }
   };
 
+  const cancelDisplayTimer = () => {
+    clearCanvasMessageTimer();
+    manifestDispatch({ autoAdvance: false, type: "setAutoAdvance" });
+  };
+
   /**
    * Switch player when navigating across canvases
    * @param {Number} index canvas index to be loaded into the player
@@ -443,6 +448,7 @@ const MediaPlayer = ({
           placeholderText={playerConfig.error}
           renderingFiles={renderingFiles}
           enableFileDownload={enableFileDownload}
+          cancelAutoAdvance={cancelDisplayTimer}
           options={options}
         />
       </div>
--- a/src/components/MediaPlayer/MediaPlayer.js
+++ b/src/components/MediaPlayer/MediaPlayer.js
@@ -293,7 +293,7 @@ const MediaPlayer = ({
    * @param {String} focusElement element to be focused within the player when using
    * next or previous buttons with keyboard
    */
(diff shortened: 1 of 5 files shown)


src/components/MediaPlayer/MediaPlayer.js
```diff
@@ -290,7 +290,7 @@ const MediaPlayer = ({
    * @param {String} focusElement element to be focused within the player when using
    * next or previous buttons with keyboard
    */
-  const switchPlayer = (index, fromStart, checkAutoAdvance = false, focusElement = '') => {
+  const switchPlayer = (index, fromStart, focusElement = '', checkAutoAdvance = false) => {
```
cjcolvar, line 293: Why did you switch these params?

cjcolvar (commented):
Looks good overall as far as I can tell.

If auto advance is turned off then should the timer countdown be hidden instead of dimmed?

src/components/MediaPlayer/MediaPlayer.js
Dananji: This was to experiment whether I could get the previous/next button focus retained in the player controls after using the previous/next buttons in the inaccessible message display, which I couldn't. This change doesn't make a difference I guess. But I can check again.

Dananji:
I did think to hide the timer instead of dimming it and with the flexbox CSS it seemed to shift the content up and down when hiding and showing the timer. And I thought that behavior wouldn't be desirable so opted to dim the timer. But hiding the timer is a better option I can explore more on that.

cjcolvar:
What do you think of putting the countdown below the previous/next buttons?  If it were below then there shouldn't be shifting from hiding/showing, right?

Dananji:
Using `visibility: hidden;` instead of `display: none;` seems to fix the issue with shifting elements.

And I removed `checkAutoAdvance` param in `switchPlayer()` function as it is not used in the current code. This flag seemed to be used when we had `switchPlayer()` triggered within the custom VideoJS progress component. And this was removed recently with the Video.js refactor.

cjcolvar (approved):
Looks great!

Issue #530, opened 2024-06-20
[BUG] Custom start not working in iOS native player


Dananji:
### Description

When there's a custom start time for the media, it seems iOS native player is not picking it up. Native player always starts from zero.

This was found in the demo site sample manifest, where it had `start` property at the Manifest-level to set the start time of playback.

Related cookbook recipe to test: https://iiif.io/api/cookbook/recipe/0015-start/

### Done Looks Like
- [ ] iOS native player start playback at given start time when there's a start time specified for the Canvas

elynema:
This should be tested in playlists. If it causes problems with playlists, we should tag this as part of 7.8 release.

elynema:
If you use a section level URL in Avalon and then add a timestamp to it, users will see an issue in native player. Avalon doesn't currently generate any URLs like this.

Playlists are working fine.

elynema:
This needs proper testing, so wait until next release, rather than rush out this release.

joncameron:
Working in iOS 16.5 Safari 👍

Issue #516, opened 2024-06-07
Scrubbing doesn't work in native player on demo site


elynema:
## Description

When testing the demo site in the iOS native player, scrubbing using the progress bar and the forward / backward 10 seconds buttons is not working at all. When the progress bar is dragged, the bar moves, but the video doesn't actually advance. When you click the forward / backward 10 seconds buttons, nothing happens to the video; it just continues playing.

I tested this in iPhone in Browserstack iOS 17 on the Ramp demo site, as well as iOS 16 on my local device. Scrubbing in the native player is working on mco-staging and on avalon-dev, so this must have been recently introduced in a Ramp change. 

Could this be related to fixes for the slow scrubbing in #494?

### Example from iOS 17 on Ramp demo site
![Jun-11-2024 10-14-03](https://github.com/samvera-labs/ramp/assets/1406618/add992cf-6468-4e76-a20e-35b199fa43e7)

## Done Looks Like
- [ ] Seeking and scrubbing is functional in iOS the native player

joncameron:
Odd rendering of VTT as well—text is left-aligned with an unexpected background color. Is this related to the same problem? If not we can create a new ticket to fix.

elynema:
Scrubbing does now seem to be working in the native player on the demo site. It is no longer working on avalon-dev, so leaving this here to finish QA once this Ramp build is pulled into Avalon.

elynema:
@charumitraravi Can you QA this?

charumitraravi:
QA completed. Scrubber is working as expected on the native player

PR #531, opened 2024-06-20
Native player

Dananji:
Related issue: #516 

Removing custom start property in the demo site manifests, since the iOS native player cannot seem to handle the custom start for the demo data. 
New issue to look at this: https://github.com/samvera-labs/ramp/issues/530


cjcolvar (approved):
Looks fine to me.  It would be good to get the start position fixed so it can be shown as an example on the demo page again.

Issue #502, opened 2024-05-20
Bring IIIF Content Search into Transcript Component


joncameron:
## Description
Intermediate functionality is needed to convert our Content Search API-based search responses to the format implemented by Third Wave in the Transcript component. For this issue, transcript component architecture isn't as important since the bulk of this work will be on parsing. Discussion of transcript component architecture can come after this work.

## Done Looks Like
- [x] Parse the search service in the manifest 
- [x] Pass search service into the search as a Matcher Factory when search functionality is called
- [x] Convert IIIF content search response into format that can be parsed by the search within component
- [x] Transcript search returns search results for all existing transcripts for current canvas.
- [x]  Drop-down to select transcript shows # of hits in each transcript (like Aviary)
- [x]  Searches submitted from the transcript search box on a media item should be submitted as phrase searches and only produce phrase matches in the transcript(s).

joncameron:
## Description
The way search is written, we may need intermediate functionality to convert search responses to the format implemented by Third Wave in the Transcript component.

There are 2 ways to use the API: to use the props, and to pass the search service in the parser and pass that on. We probably would go with the latter (parsing), and we'll need a ticket to parse the search service in the manifest and pass that into the search as a Matcher Factory to call it when searching is happening in the UI.

joncameron:
Other thing: does the Transcript component need to be independent still? If so, we may need to change code; Patrick has pulled the component under the overall state which will complicate standalone usage. We should have a discussion on this: what are our needs? Do we need to keep it any particular way?

joncameron:
See samvera-labs/ramp#502 for discussion of retaining independence for the Transcript component.

elynema:
@Dananji I'm seeing an issue with this example record: https://avalon-dev.dlib.indiana.edu/media_objects/gf06g267d. I don't know if it's related to re-formatting the IIIF content search or if it's related to displaying search results.

For the first section, if I go to the transcript tab and search 'Isaac' I get no results, even though Isaac is clearly in the first few lines of the transcript. The underlying IIIF content search is providing a result, but it's not getting shown in the front-end.

elynema:
Same issue when I search 'depaul' or 'university' in this transcript. Hits are not being shown in the interface. I am not seeing this issue in other sections on this record so not sure if it is related to the transcript or section in particular.

elynema:
[...]

PR #532, opened 2024-06-21
Adjust content search when switching between transcripts

Dananji:
Refinements to feature added in #502 

This PR includes the following refinements to content search:
- Use cached search results when switching between transcripts in the same Canvas without changing the query
- Preserve search counts when switching to a transcript where search is disabled
- Update search results counts properly when switching between transcripts
- Clear search counts when current search is cleared
- Display hit counts correctly when there are multiple hits in a single cue
- General refactoring to parts of code related to transcript component

### DIFF
--- a/src/services/transcript-parser.js
+++ b/src/services/transcript-parser.js
@@ -10,6 +10,7 @@ import {
   identifyMachineGen,
   identifySupplementingAnnotation,
   parseSequences,
+  groupBy,
 } from './utility-helpers';
 import { getCanvasId } from './iiif-parser';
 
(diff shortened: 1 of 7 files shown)


src/services/transcript-parser.js
```diff
@@ -827,37 +832,55 @@ export const getMatchedTranscriptLines = (searchHits, query, transcripts) => {
...
+ * @returns matched cue with HTML tags added for marking the hightlight 
  */
-export const getMatchedParts = (offset, text, query) => {
-  return [
-    text.slice(0, offset),
-    text.slice(offset, offset + query.length),
-    text.slice(offset + query.length)
-  ];
+export const markMatchedParts = (text, query) => {
+  const queryRegex = new RegExp(String.raw`${query}`, 'gi');
```
cjcolvar, line 860: This appears to be too broad a regex.  When I search for 'you' on the first section of https://avalon-dev.dlib.indiana.edu/media_objects/g158bh28p/manifest.json, I get 35 hits but 55 highlights.  From looking at the content search results it looks like 35 is correct.  Some examples of false highlights include `your`, `you'd` and `you'll`).

cjcolvar (commented):
I did a little testing and search feels improved.  I ran into a couple inconsistencies:
- highlighting words within other words (e.g. `you` and `your`)
- phrase searching correctly highlights but hit count is 0 (e.g. `feeling better`)

src/services/transcript-parser.js
Dananji: Good catch! I'll take a look at it.

src/services/transcript-parser.js
```diff
@@ -827,37 +832,76 @@ export const getMatchedTranscriptLines = (searchHits, query, transcripts) => {
     }
     const matchOffset = mappedText.toLocaleLowerCase().indexOf(qStr);
     if (matchOffset !== -1 && transcirptId != undefined) {
```
masaball, line 834:
```suggestion
    if (matchOffset !== -1 && transcriptId != undefined) {
```

Not a blocker, but all mentions of `transcriptId` in this function are typoed. It is still all working because the typo is consistent, but would probably be good to clean up.

masaball (approved):
As far as I can tell this looks good.

Issue #473, opened 2024-04-04
Unusual Behavior For Captions in iOS Native Player


joncameron:
# Description

# To Reproduce
- Go to https://avalon-dev.dlib.indiana.edu/media_objects/cj82k730p in iOS 17 on Browserstack
- Turn off captions
- Select the Play button
- Captions automatically turn on again and the player goes into native mode fullscreen
- Select the caption labeled as French from the Subtitles menu in the native player
- No captions display in the player
- Select the caption labeled as English
- Captions display in the player
- Select the caption labeled as French  
- Captions now display in the player

Alternatively, to reproduce
 - go to: https://avalon-dev.dlib.indiana.edu/media_objects/rr171x25z in iOS 17
 - play video
 - CC option is selected by default in the native player
 - select off in the subtitle menu (should turn off, but sometimes that doesn't work)
 - select English option in the subtitle menu
 - captions do not display in player
 - if I go back and try off again, and come back to english option, I have been able to get captions to come back on

# Done Looks Like
- [ ] Investigation into why the native iOS player is out of sync with Video.js
- [ ] Create known issues for Avalon with turning captions off/on in native player vs. web player
- [ ] Create known issues for Ramp with turning captions off/on in native player vs. web player
- [ ] remember prior selected text track in web view player when returning from native player

joncameron:
Changes made in the video.js player seem to be replicated in the native player but it's broken in this particular instance.

Dananji:
In the second scenario in the description;

> 
> - go to: https://avalon-dev.dlib.indiana.edu/media_objects/rr171x25z in iOS 17
> - play video
> - CC option is selected by default in the native player
> - select off in the subtitle menu (should turn off, but sometimes that doesn't work)
> - select English option in the subtitle menu
> - captions do not display in player
> - if I go back and try off again, and come back to english option, I have been able to get captions to come back on
> 

From what I've experienced, when captions are turned off in the native player for HLS media, there's a set of specific steps to follow to enable them again. This was an issue we had with the previous MediaElement.js implementation as well.

Once captions are turned off in the native player, to turn them back on click `CC` option and then the captions options to be turned on. And this example `English` option.

Or this can be turned on the Video.js player after exiting the native player by just clicking on the `English` option in the captions menu.

elynema:
From testing the first case listed above:

- Loaded https://avalon-dev.dlib.indiana.edu/media_objects/cj82k730p and turned captions off
- Played video.
[...]

PR #534, opened 2024-06-25
Preserve captions off state in native iOS player on play event

Dananji:
Related issue: #473


cjcolvar (approved):

Issue #537, opened 2024-06-28
Update vtt / srt validation for hours


elynema:
# Description

Technically, vtt and srt file formats are required to either have no digits for the hour in the timestamp, or 2 digits. Currently the validation in the Ramp transcript component allows 1 digit in the hour in the timestamp.

# Done Looks Like

 - [x] Ramp shows an error if a vtt or srt transcript has a single digit for the hour

masaball:
[amp-demo-amia-transcript (2).txt](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK2VlQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--6896494d7062ff428205c17e7d5861bdf5a7d5c8/amp-demo-amia-transcript%20%282%29.txt)

joncameron:
https://avalon-dev.dlib.indiana.edu/media_objects/bc386j28j

PR #546, opened 2024-07-01
SRT/VTT timestamp validation to deny single digit hour in hh:mm...

Dananji:
Related issue: #537 

Display a message when at least one of the timestamps in SRT/VTT cues are invalid:

Screenshot 2024-07-01 at 3 34 43 PM

### DIFF
--- a/src/services/transcript-parser.js
+++ b/src/services/transcript-parser.js
@@ -33,7 +33,8 @@ const TRANSCRIPT_MIME_EXTENSIONS = [
 
 // ENum for describing transcript types include invalid and no transcript info
 export const TRANSCRIPT_TYPES = {
-  invalidTimedText: -3,
+  invalidTimestamp: -4,
+  invalidVTT: -3,
   noSupport: -2,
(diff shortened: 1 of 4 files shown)


src/services/transcript-parser.js
```diff
@@ -566,14 +567,22 @@ export function parseTimedText(fileData, isSRT = false) {
   const groups = groupTimedTextLines(cueLines);
   // Add back the NOTE(s) in the header block
   groups.unshift(...noteLines);
+  let hasInvalidTimestamp = false;
   groups.map((t) => {
     let line = parseTimedTextLine(t, isSRT);
     if (line) {
       tData.push(line);
+    } else {
+      hasInvalidTimestamp ||= true;
```
cjcolvar, line 576: What about breaking the loop here since the error case has already been hit?

cjcolvar (commented):
Looks good.  I just had one small suggestion.

cjcolvar (approved):

Issue #538, opened 2024-06-28
In Safari, transcript search result paging is on same line as search box, rather than below


elynema:
# Description

In Safari on desktop, I am seeing the paging for the transcripts search results squeezed onto the same line as the search box. This is making the search box really small. It does not matter how wide I make my browser.

<img width="1729" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBM3FmQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--6da7cabd6f285324c00a3f7f7b7052f1c3c87dc6/image.png" alt="image.png" />

In Chrome on desktop, the paging is on a line below the search box. This seems like a better fit as it doesn't squish the search box.

<img width="1402" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMzJmQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b80f6e9d8956bc980f04f5f8258508defe89d7bb/image.png" alt="image.png" />

# Done Looks Like

 - [ ] Safari layout for search result paging is the same as Chrome

elynema:
@Dananji I assume the way that Chrome displays this is the way that we originally intended? If Safari is displaying as expected, perhaps we should discuss.

Dananji:
This is a grid display which is responsive to the screen size. I was not able to reproduce the layout in Safari in my machine (probably because the screen size is larger).
I think the view that makes most sense it the one on Chrome, which allows space for the search input field to expand the full width of the container.

joncameron:
The pager is now appearing on a separate line in Safari as expected.

PR #547, opened 2024-07-02
Use flexbox instead of grid to arrange transcript menu

Dananji:
Related issue: #538


cjcolvar (approved):

PR #550, opened 2024-07-03
Update setTimeout in the debounce

Dananji:
Related issue: #540 

I removed the 5 millisecond timeout in the `setTimeout` and now it runs with a zero delay. This enables the browser to schedule these requests without immediately executing them. And it helps to avoid any unnecessary UI updates for intermediate responses.

I checked time it takes for the Ramp to process the search response (UI to updates right after this).

And I did this test for multiple transcripts in different items and I was seeing numbers less than 50 milliseconds (0.05 seconds) each time.


cjcolvar (approved):

Issue #513, opened 2024-05-31
Untimed Text Transcripts in Transcript Search Component


joncameron:
## Description
Untimed text files should be able to be searched just like timed text transcripts, but the search component developed by Third Wave doesn't account for untimed text material. Ramp should be able to support search and highlighting for these transcripts as well. Currently, results returned from the search service for untimed text aren't set up to be highlighted and loaded into the search results for navigation.

The current JS implementation by Third Wave may also not be able to index and query Word Docs (not designed for untimed text).

The tricky part in this implementation is figuring out how to use the search hits in non-timed text to work with previous/next button in the results navigator since they are not indexed in the JS code. Highlighting the search hits in the transcript display shouldn't be that hard as we could re-use the highlights in the search response to do this.

This can most likely wait until after 7.8 release.

## Done Looks Like
 - [ ] for txt or docx transcripts, users can enter search term(s) on the Transcript tab and get a set of results to iterate through
 - [ ] results should be highlighted by paragraph (at least), but each paragraph may have multiple hits
 - [ ] result counts from txt or docx transcripts not currently displayed should show up in the drop-down
 - [ ] since there is no auto-scroll for untimed text, users can page through transcript search hits while the media item is playing

elynema:
Note that the lack of this feature can be somewhat confusing. For instance, a hit may be returned in search results due to a match in the .txt transcript, but the hit cannot be found using the search within feature on the media object page.

For example, search 'basketball' in the repository. 
One of the hits is: https://avalon-dev.dlib.indiana.edu/media_objects/x920fw86z

It has 3 transcripts. The first 2 webvtts are searchable, but do not contain the word basketball. The last txt transcript does, but it does not show a hit count next to it and there is no search within feature when you select that transcript.

This is potentially confusing for users.

elynema:
Another example where the hit gets pulled up as a search result for transcript hits if you search 'indianapolis', but you can't actually search the 2 transcripts available: https://avalon-dev.dlib.indiana.edu/media_objects/9k41zd49s

elynema:
For QA, we should test a variety of formatting:

- paragraphs separated by one newline instead of two
- vtt that is saved as a txt file

Check that count and navigation is working properly.

Dananji:
This can be tested on Ramp demo

elynema:
@Dananji I took a first pass on the demo site.

When I type in >1 search term, the hits are italicized, but not bolded with color change in the transcript text.

[...]

PR #558, opened 2024-07-09
Text search nav

Dananji:
Related issue: #513 

This PR adds the ability to see focused search hits using results navigation when there are multiple hits within a focused cue.

Example of search results navigation when the transcript is one block of text:
Screenshot 2024-07-09 at 5 11 46 PM

### DIFF
--- a/src/services/search.js
+++ b/src/services/search.js
@@ -170,7 +170,7 @@ export function useFilteredTranscripts({
   }, [matcher, query, enabled, sorter, matchesOnly, showMarkers, playerDispatch, selectedTranscript]);
 
   const callSearchFactory = () => {
-    clearTimeout(debounceTimerRef.current);
+    if (!debounceTimerRef.current) clearTimeout(debounceTimerRef.current);
 
     const abortController = new AbortController();
     abortControllerRef.current = abortController;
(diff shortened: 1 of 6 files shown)


src/services/utility-helpers.js
```diff
@@ -515,6 +507,15 @@ export function getLabelValue(label) {
...
 
+export function decodeHTML(labelText) {
+  return labelText
+    .replace(/&amp;/g, '&')
+    .replace(/&lt;/g, '<')
+    .replace(/&gt;/g, '>')
+    .replace(/&quot;/g, '"')
+    .replace(/&apos;/g, "'");
+};
+
```
cjcolvar, line 518: It looks like we already have libraries (`entities`, `html-entities`, `he`, and `parse-entities`) installed that do html entity encoding/decoding.  What about replacing this with one of those or is a more limited decoding preferable?

Dananji: Replacing this with one of these sounds good to me.

src/services/search.js
```diff
@@ -343,5 +343,11 @@ export const useFocusedMatch = ({ searchResults }) => {
     }
   }, [searchResults.matchingIds, focusedMatchIndex]);
 
+  useEffect(() => {
+    if (searchResults.matchingIds.length && focusedMatchIndex > 0) {
+      setFocusedMatchIndex(null);
```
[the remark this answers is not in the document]
Dananji: Yes, I see now that `null` is used to indicate disabled state. You are correct, setting it to -1 seems like the best solution. Good catch!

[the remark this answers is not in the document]
Dananji: Setting this to -1 reset the index to 0 when switching between transcripts in the same canvas after navigating through hits. e.g. when changing transcripts with the results nav saying `2 of x` set the index to zero for the new transcript (`0 of y`).
So I didn't change it. For some reason the `null` works in this.

src/services/utility-helpers.js
Dananji: Replace it with `decode` from `html-entities`.

cjcolvar (approved):
This looks good as far as I can tell.

Issue #559, opened 2024-07-10
Audio item volume slider out of position in some contexts


masaball:
The audio slider position circle is out of position; it seems to be a little lower than expected and not centered on the slider/rail in the Avalon embedded context.

# Screenshots
<img width="603" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeHpGQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--28567641892565b80305e242ffe422f43e027498/Screenshot%202024-07-09%20at%2010.09.45%E2%80%AFAM.png" alt="Screenshot 2024-07-09 at 10.09.45 AM.png" />

<img width="606" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeDNGQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0cfaa7aa576f9dca7e2bc71ce32076b45b168a6e/Screenshot%202024-07-09%20at%2010.09.36%E2%80%AFAM.png" alt="Screenshot 2024-07-09 at 10.09.36 AM.png" />

## Done Looks Like
- [ ] Volume slider nub is centered on the volume slider rail in all contexts

joncameron:
Looks good!
<img width="772" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL25YQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--8d25f043f2ed6889bc96676c853d64168c19715d/Screenshot%202024-07-12%20at%201.30.41%E2%80%AFPM.png" alt="Screenshot 2024-07-12 at 1.30.41 PM.png" />

PR #562, opened 2024-07-10
Small tweak to top margin of horizontal volume slider thumb

cjcolvar:
Before:
Screenshot from 2024-07-10 17-16-19

After:
Screenshot from 2024-07-10 17-15-15

Part of #559

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -83,7 +83,7 @@
 }
 
 .vjs-slider-horizontal .vjs-volume-level .vjs-svg-icon {
-  margin-top: 0.1em;
+  margin-top: 0.15em;
 }
 
 /* Make player height minimum to the controls height so when we hide


Dananji (commented):
It looks good from the attached screenshot in the PR.

But, in the ticket the screenshot shows that the thumb is below the volume slider bar unlike the before screenshot in the PR. 
So, my guess is that increasing `margin-top` wouldn't work in the Avalon context, since this change pushes the thumb further down.
Other possible solutions:
- try using a different CSS property to align things properly (like `vertical-align` on the SVG)?
- tweak this in Avalon code if there's no other way to fix this on the Ramp side?

What do you think @cjcolvar ?

Dananji (approved):
Sorry, I didn't see that you had made the change in Avalon repo before. So, I guess this is good for Ramp side of things.

Issue #548, opened 2024-07-02
Media Player Has Gray Overlay When Moving Between Items


joncameron:
# Description
If you start from a playable playlist item that immediately follows an inaccessible item and click previous canvas, the inaccessible item doesn't fully load. It displays with timer, but is grayed out looking and the previous and next buttons are disabled. To reproduce, load the playlist, click on playlist item #8 and click the previous canvas button. Note that after I've played with this playlist a bunch, the problem is not always reproducible unless I reload the playlist.

## Related bugs

- Select playlist item 17, and hit the next canvas button, playlist item 18 never loads unless I select it from the structured navigation. 
- Select playlist item 24, I can use next canvas to advance to 25, but then the next canvas button does nothing (won't advance to 26).
- In Safari on iPad, if I advance to or select item 2, the next/previous canvas button will not return to me to item 1 (an inaccessible item) or item 3 (an accessible item). This was not observed on desktop.
- In Safari on iPad, if I select 29, the previous and next canvas buttons do not work (nothing happens when I click them). This was not observed in Safari on desktop (worked properly). 
- If I use the next canvas button to advance to 29, the previous canvas button skips 28 (an inaccessible item). Observed in Safari on desktop and iPad. In Chrome on Android, the previous canvas button returns to 8 as it should, but sometimes there is a gray overlay.

## Example
https://avalon-dev.dlib.indiana.edu/playlists/5

# Done Looks Like
- [x] Gray overlay doesn't stick around or block UI when moving between playlist items
- [x] Test transition from playable item to inaccessible item
- [x] Test transition from inaccessible item to playable item
- [x] Next/Previous canvas buttons work when clicked
- [x] When using previous/next canvas buttons the previous/next items fully load;
  - [x] a playable canvas shows player controls and can be played
  - [x] an inaccessible items shows the previous/next buttons with a timer (when autoplay is on) without a grey overlay

Dananji:
Created a new issue for the mobile devices: https://github.com/samvera-labs/ramp/issues/552

elynema:
@Dananji I tested the 4 cases listed above that mentioned desktop in Safari and Chrome.

I ran into 2 issues that were intermittent, unfortunately. Perhaps you can see if you are able to reproduce them? 

[...]

PR #565, opened 2024-07-11
Fix grey overlay bug and related bugs in playlist item transitions

Dananji:
Related issue: #548


cjcolvar (approved):
Looks good as far as I can tell.  Hope these small timing tweaks make things better overall!

Issue #564, opened 2024-07-11
[BUG] Results from content search not displayed or highlighted in transcript when they include non-alphanumeric characters


elynema:
**Describe the bug**
When the term(s) entered into the transcript search box container certain non-alphanumeric characters, the transcript search says 'no results found' and does not highlight results in the transcript, even though hits are being returned from the content search API itself.

**To Reproduce**
1. Go to the transcript for the first section: https://avalon-dev.dlib.indiana.edu/media_objects/gf06g267d
2. Enter one of the following searches into the transcript search box:
3. [THEME MUSIC]
4. --Television 
5. J-U-N-G
6. "Dear
7. The transcript search says 'no results found' and no results are highlighted in the transcript.

**Expected behavior**
If there are results returned from the content search response, they should be displayed as results in the transcript component. There will be times when the content search strips some of these characters from its search, so there are responses that do not contain them. Examples are the '--television' search. Will that be a problem for displaying those as a match?

**Screenshots**
<img width="1402" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNnZSQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e4d49eb1bcf63e9bea1062703551decbb7ab10a2/image.png" alt="image.png" />

**Desktop (please complete the following information):**
 - OS: Mac
 - Browser: Chrome and Safari
 
**Additional context**
This was discovered during testing for #5897 in avalon repo, so more details are available there.

joncameron:
These are now matching as expected.

PR #566, opened 2024-07-11
Fix regex to identify matches in content search response

Dananji:
Related issue: #564

### DIFF
--- a/src/services/transcript-parser.js
+++ b/src/services/transcript-parser.js
@@ -831,7 +831,6 @@ export const getMatchedTranscriptLines = (searchHits, query, transcripts) => {
   if (searchHits === undefined) return;
 
   let traversedIds = [];
-  const cleanedQuery = buildRegexReadyText(query);
 
   searchHits.map((item, index) => {
     let { target, value } = item;
@@ -854,7 +853,7 @@ export const getMatchedTranscriptLines = (searchHits, query, transcripts) => {
       const matchOffset = mappedText.toLocaleLowerCase().indexOf(queryText);
 
       if (matchOffset !== -1 && transcriptId != undefined) {
-        const match = markMatchedParts(value, cleanedQuery, qStr, item.hitCount, true);
+        const match = markMatchedParts(value, qStr, item.hitCount, true);
 
         transcriptLines.push({
           tag: TRANSCRIPT_CUE_TYPES.timedCue,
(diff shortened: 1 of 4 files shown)


cjcolvar (commented):
I tried testing this locally by converting the Gatsby test data into a VTT ([attached as Gatsby.vtt.txt](https://github.com/user-attachments/files/16254327/Gatsby.vtt.txt) since I had to change the extension to get github to attach it), loading that into Avalon, then testing within ramp.  Searching worked as expected for all of the test cases in this PR, but I found a couple issues:
- Searching 'invited' only shows one result in ramp when the content search is returning two (the second being '<em>invited</em>-they')
- Searching 'dr eckleburg who' or 'went there they' will match correctly but remove the remainder of the cue text

It seems like Ramp is trying really hard to mimic the behavior of Avalon's content search and is doing a pretty good job at it, but I'm afraid of Ramp having to keep up with changes if Avalon changes how it's content search works.  I'm wondering if instead of trying to match exactly like how the content search matches Ramp tries to replace the match text entirely with what comes back from the content search and then adjusts the match tags as it sees fit.  This approach relies more on the content search which will probably have it's own issues.

If the two issues I saw can be easily fixed then I think this implementation might be good enough for now.  If we run into too many edge cases then we may want to consider trying the replacement approach, but I'm hoping that can wait until after Avalon 7.8.

src/services/transcript-parser.js
```diff
@@ -1006,14 +1006,19 @@ export const markMatchedParts = (text, query, hitCount, hasHighlight = false) =>
     let startIndex = 0;
     let newStr = '';
     for (let j = 0; j < matches.length && count < hitCount;) {
-      if (matches[j] === undefined && matches[j + 3] === undefined) return;
+      if (matches[j] === undefined && matches[j + offset] === undefined) return;
+
+      // Set offset to count matches based on the # of words in the phrase search query
+      const offset = query.split(/[\s-,\?]/)?.length > 0
+        ? (query.split(' ')?.length) * 2 - 1 : 1;
```
cjcolvar, line 1013: Is it a problem that `offset` isn't defined until after it is used on line 1009?  Could these two lines be switched?

Dananji: Yes, absolutely!

cjcolvar (approved):
I retested and the two issues are resolved for me! :tada: 
I am seeing console logging when a query has a single character after a space.  It is still matching and highlighting when there is a valid match like "night I" so I'm not sure if this is really a problem or not.

```
Error building RegExp for query:  night i
Error building RegExp for query:  who h
Error building RegExp for query:  automobiles w
Error building RegExp for query:  automobiles which b
```

Dananji:
These console logs are related to the error I mentioned where, the function building the regular expression bombing when it is called with intermediate queries. 

This doesn't impact search within functionality as far as I know. It is just the handling of intermediate states of the search query when it is typed in.

Issue #551, opened 2024-07-05
Selecting Playlist Item from List Changes Autoplay Behavior


joncameron:
# Description
When a user manually selects any playlist item in the navigation, it kind of turns auto-play off (not the toggle, just the way that the playlist behaves). If I select a playlist item, and then start playing it, the next playlist item does not play automatically. It has to be manually played. The next inaccessible item will count down, but accessible content will not actually play and the following item in the playlist does not load and start playing automatically as would be expected.

# Done Looks Like 
- [x] Selecting a playlist item does not affect the autoadvance and play behavior when autoplay is enabled

elynema:
@joncameron Just noting that this issue does seem to be a regression from 7.7.2. Did you want to pull it into 7.8 or no?

joncameron:
I think it would be good to fix this—added to 7.8.

joncameron:
Could no longer reproduce.

PR #569, opened 2024-07-15
Ensure autoplay of next section when selecting inaccesible item

masaball:
Related issue: #551


cjcolvar (approved):

PR #571, opened 2024-07-17
Fix slow search by handling timing of updates and cache

Dananji:
I was able to reduce the UI update time by ~2 seconds (~4 seconds to ~1 second) for search within functionality from the fixes made in this PR. 

I think this could be further improved by restricting the updates to props that goes into `TranscriptList` component with search within, maybe?


cjcolvar (approved):
Looks good as far as I can tell. Nice work speeding things up! :rocket:

PR #574, opened 2024-07-17
Change vertical-align CSS for SVG

Dananji:
Related issue: https://github.com/avalonmediasystem/avalon/issues/5927

Once this is brought into Avalon we probably could undo the changes in https://github.com/avalonmediasystem/avalon/pull/5921 as this takes care of the misalignment of svg icons in Ramp.


cjcolvar (approved):
Nice!  Glad this works and doesn't disrupt anything else.

PR #576, opened 2024-07-18
Add `withCredentials` option to MediaPlayer component

### LINKED ISSUE #575: Can't access cross-origin video that requires cookie auth
**Describe the bug**

We’re using CloudFront streaming with signed cookie authentication. I can see the cookies being set correctly when the page is loaded, but those cookies are not being sent to the streaming server by VideoJS. I managed to brute-force a working scenario, but I don't have enough knowledge of the Ramp project to know how best to integrate it as an option.

**To Reproduce**
1. Set up a video streaming server that returns a `403` unless the proper cookie is included with the request
2. Set up a server on a different host in the same domain that will serve an HTML page with a Ramp `MediaPlayer` object in it, pointing at the video streaming server from Step 1
3. Include the proper `Set-Cookie` response header in that HTML page response
4. Load the page
5. See the `403` error from the video streaming server

**Expected behavior**
The `MediaPlayer` object (optionally) includes the proper configuration to include credentials with cross-origin requests.

**Screenshots**
set_cookie
Note the `Set-Cookie` headers returned with the page, with the domain set up one level to allow the browser to send them to the streaming server in the same subdomain.

no_cookies
Note that the cookies are _not_ included with XHR requests to the `httpstream` host, despite having a subdomain match. The video is not loading (as you can see from the time indicator).

cookies
After adding `hls: { withCredentials: true },` right above [this line](https://github.com/samvera-labs/ramp/blob/main/src/components/MediaPlayer/MediaPlayer.js#L374), the cookies were sent and the video loaded correctly.

**Desktop (please complete the following information):**
 - OS: MacOS 14.5 (Sonoma)
 - Browser Firefox
 - Version 128.0

**Additional context**
This bug is a major blocker in our attempt to get our streaming server upgraded to the latest release of Avalon.

mbklein:
Resolves #575 

This PR adds a `withCredentials` boolean property to the `MediaPlayer` component to allow the underlying VideoJS component to include cookies and auth info in its XHR requests. Requires that the server be configured to include the correct CORS headers to allow the browser to send credentials. These requirements are included in the documentation of the new property.


Dananji (approved):
This looks good 👍 

Thank you for the PR Michael!

Issue #567, opened 2024-07-12
Progress bar shortened on initial manifest load


masaball:
**Describe the bug**
When Ramp first loads, the progress bar is being heavily truncated. This is being seen on media objects and playlists.

The progress bar fixes itself/fills properly on canvas switching and does not revert to the shortened form if you return to the original canvas.

**To Reproduce**
On Avalon-dev:
1. Go to any media object or playlist show page
2. When the page loads, observe the shortened progress bar
3. If the object has structure, click on a sub-section
4. The light gray does not change size, but the highlighting will be over the correct part of the full length allotted
5. The playhead will move to position on the light gray part of the progress bar
6. If there are multiple items, select a different item
7. Observe that the progress bar fills properly

On Ramp demo site:
1. Load the ramp demo site: https://ramp.avalonmediasystem.org/
2. Observe the incorrect progress bar
3. Copy a manifest link from avalon-dev: https://avalon-dev.dlib.indiana.edu/media_objects/g158bh28p/manifest.json
4. Paste it into the "Manifest URL" box and click "Set Manifest"
5. Once page loads, observe the incorrect progress bar
6. Behavior of highlighting sections and switching canvases is the same as the Avalon-Dev case

**Expected behavior**
The light gray progress bar extends the full width allotted

**Screenshots**
<img width="611" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN2pZQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--8f80acb9dddaad7122b690e2918040b5a731daf2/Screenshot_20240712_160854.png" alt="Screenshot_20240712_160854.png" />

<img width="600" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN25ZQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f72b4f843316afd6dadcba6f0872679b84ee2650/Screenshot_20240712_162637.png" alt="Screenshot_20240712_162637.png" />

**Desktop (please complete the following information):**
 - OS: Ubuntu 22.04
 - Browser: Firefox v.128 and Chrome v.126.0.6478.126

joncameron:
[...]

PR #578, opened 2024-07-18
Read duration from either options or player property in progressbar

Dananji:
I was still seeing the shortened progressbar on initial load in the Ramp demo site (#567) in Firefox. It seems on the initial load of the page it's safer to read the duration from options passed into the `VideoJSProgress` component rather than relying on `canvasDuration` property of the player instance.

### DIFF
--- a/src/components/MediaPlayer/MediaPlayer.js
+++ b/src/components/MediaPlayer/MediaPlayer.js
@@ -315,7 +315,7 @@ const MediaPlayer = ({
   };
 
   React.useEffect(() => {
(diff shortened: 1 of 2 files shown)


cjcolvar (approved):
:+1:

Dananji:
This was still being buggy, so I removed the `duration` prop to progress bar component since that prop stays the same in subsequent player refreshes. Now the progress bar component relies on the `player.canvasDuration` property which gets updated on canvas change.

cjcolvar (approved):
Looks good!

Issue #580, opened 2024-07-22
[BUG] Auto-play does not start next item in playlist if user clicks next canvas to inaccessible item


elynema:
**Describe the bug**
If the user selects an item that is immediately before an inaccessible item, starts it playing, then hits 'next canvas' to the inaccessible item, and then allows the inaccessible item to expire, the next playlist item does not automatically begin playing.

This is an edge case remaining from testing behavior around auto-play in playlists with the new prevous/next button functionality for inaccessible playlist items. 

**To Reproduce**
1. Go to avalon-dev, do not log in.
2. Go to playlist: https://avalon-dev.dlib.indiana.edu/playlists/5
2. Make sure auto-play is turned on
3. Select playlist item 18. Hit the play button to start playback
4. Before item 18 is finished playing, hit the next canvas button.
5. Playback proceeds to the inaccessible playlist item 19. 
6. Let the 10 seconds expire
7. Playback proceeds to the accessible playlist item 20.
8. However, playback of playlist item 20 does not begin automatically.

**Expected behavior**
Since auto-play is turned on, playback of any accessible playlist item should begin automatically.

Note: Sometimes this works as expected the first time attempted after the playlist page loads. If so, try again. I have found that the second try it stops working. This can also be tested on any single inaccessible playlist item (there are several in this playlist).

**Desktop (please complete the following information):**
 - OS: Mac laptop
 - Browser: Chrome and Safari

Dananji:
Fixed behavior in PR:
- If the user has started playback at some point play the next item after an inaccessible item when the timer ends
- If the user has not started playback at any point/started playback and paused it before reaching the inaccessible item don't play the next item after an inaccessible item when the timer ends

elynema:
This is probably a regression of us re-organizing the way that video.js initializes in 7.8.

joncameron:
Can no longer reproduce.

PR #581, opened 2024-07-23
After an inaccessible item play the next item if user has started playback

Dananji:
Related issue: #580 

Fixed behavior in PR:
- If the user has started playback at some point play the next item after an inaccessible item when the timer ends
- If the user has not started playback at any point/started playback and paused it before reaching the inaccessible item don't play the next item after an inaccessible item when the timer ends

**Not sure whether this can go into 7.8, needs to discuss before merging**

### DIFF
--- a/src/components/StructuredNavigation/StructuredNavigation.js
+++ b/src/components/StructuredNavigation/StructuredNavigation.js
@@ -131,13 +131,6 @@ const StructuredNavigation = () => {
         playerDispatch({
           type: 'resetClick',
         });
-        // Ensure autoplay from inaccessible items in playlists
-        if (playlist) {
-          playerDispatch({
-            isPlaying: autoAdvance,
-            type: 'setPlayingStatus',
-          })
-        }
       }
     }
   }, [isClicked, player]);
--- a/src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
+++ b/src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
@@ -566,11 +566,13 @@ function VideoJSPlayer({
     playerLoadedMetadata(player);
 
     player.on('pause', () => {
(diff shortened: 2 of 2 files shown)


src/components/StructuredNavigation/StructuredNavigation.js
```diff
@@ -131,13 +131,6 @@ const StructuredNavigation = () => {
         playerDispatch({
           type: 'resetClick',
         });
-        // Ensure autoplay from inaccessible items in playlists
```
cjcolvar, line 134: This is undoing https://github.com/samvera-labs/ramp/pull/569.  Did you test that case (clicking on a playlist item in a playlist disables autoplay without untoggling the switch https://github.com/samvera-labs/ramp/issues/551) is resolved without this code and your changes in VideoJSPlayer.js?

Dananji: The changes in #569 would always play the clicked (selected) item after an an inaccessible item regardless, whether the user had initiated playback at any point, which I think is undesirable behavior. Because it is similar to auto-play without any indication to the user, which violates accessibility guidelines.

And testing this with changes in this PR;
- plays the clicked item after an inaccessible item if the user has initiated playback before
- do not play the clicked item after an inaccessible item if the user has not initiated playback before reaching the inaccessible item

Does this still sound like what we need or does this behavior need to change? 

Because from the description in #551 I can't figure out whether we need to play items automatically after an inaccessible item count down ends without user initiating playback before they click on an inaccessible item.

cjcolvar: I tested this out and as far as I can tell it is working as I would expect and I'm not able to reproduce #580 or #551.

cjcolvar (approved):

Issue #560, opened 2024-07-10
[BUG] Playlist item does not playback intermittently in Safari


elynema:
**Describe the bug**
In our testing playlist, there is a playlist item that sometimes will not play when advancing to it using the next canvas button. This has been observed in Safari with the playlist on both avalon-dev and in the Ramp demo site. This happens intermittently (so not always).

It is unknown if this problem might occur for other playlist items.

**To Reproduce**
1. Go to playlist: https://avalon-dev.dlib.indiana.edu/playlists/5/manifest.json
2. Select playlist item 17, and hit the next canvas button. 
3. Playlist item 18 loads, but can't actually play (just spins if you try to play it, and the poster image doesn't load)
4. In the Ramp demo site, the playlist actually jumps ahead to playlist item 19 if auto-play is turned on but playlist item 18 can't be played.

**Expected behavior**
When you advance to playlist item 18 using the next canvas button, the poster image should load and the content should play if the user hits the play button.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Mac
 - Browser: Safari
 - Version 17.5 (17618.2.12.111.5, 17618)

Dananji:
This seems to be a timing issue with player events and the custom progress bar updates not aligning properly. I have tried fixing the timing in multiple places, and it breaks in another place. This is intermittent in behavior.

As this is not urgent for 7.8 and will probably go in to the next release and we've been having issues with timing in events in the progress updates lately that we've been patching up, I thought this would be good chance to re-examine the implementation of the custom progress bar.

So I'm trying to re-organize the Video.js components using Video.js' native progress control with some custom implementations to see whether we can overcome these timing issues.

Dananji:
I was not able to make it work with re-organizing the Video.js components using Video.js' native progress control. However using React refs for variables along with the fix for #580 this seems to be working.

joncameron:
Could no longer reproduce. 👍

PR #587, opened 2024-07-23
Update variables in progress component to refs to maintain accuracy with canvas updates

Dananji:
Related issue: #560 

~~**Merge this after 7.8 related Ramp release is created**~~

After some discussion in Slack it was decided to merge this for 7.8 release since this has similar scope as #580 which is going to be in 7.8.

### DIFF
--- a/src/components/MediaPlayer/VideoJS/components/js/VideoJSProgress.js
+++ b/src/components/MediaPlayer/VideoJS/components/js/VideoJSProgress.js
@@ -48,7 +48,8 @@ class VideoJSProgress extends vjsComponent {
   /** Build progress bar elements from the options */
   initProgressBar() {
     const { targets, srcIndex } = this.options;
-    const { start, end, duration } = targets[srcIndex];
+    const { start, end } = targets[srcIndex];
+    const duration = this.player.canvasDuration;
     let startTime = start,
       endTime = end;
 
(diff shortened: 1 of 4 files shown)


src/components/MediaPlayer/VideoJS/components/js/VideoJSProgress.js
```diff
@@ -274,17 +283,20 @@ function ProgressBar({
...
     setTLeft(left);
 
-    const curTime = player.currentTime();
-    setProgress(curTime);
-    setCurrentTime(curTime + canvasTimes.altStart);
+    setProgress(cTimes.start);
+    setInitTime(cTimes.start);
+
+    setCurrentTime(cTimes.start);
+    player.currentTime(cTimes.start);
```
cjcolvar, line 299: Can you explain these changes?  I don't understand the change of setting currentTime to `cTimes.start` instead of `player.currentTime() + cTimes.altStart`.

Dananji: The `altStart` applies to the use-case with multiple files in a single Canvas (cookbook recipe 64). I tested these changes for that recipe just now and it seems this line didn't make a difference for that use-case, so we can remove this . I guess, this is one of the code lines we used in the initial setup stage and somehow gotten overridden in another place? However I noticed one tiny change needed for timeupdate, which I will commit right after this.

cjcolvar (approved):

PR #592, opened 2024-07-24
Only reset isClicked if active structure item is set

cjcolvar:
This fixes an error I was seeing with an item in staging but not in production.  Without this fix ramp crashes before fully loading with `TypeError: k.current is null`
https://mco-staging.dlib.indiana.edu/media_objects/dj52wp041/manifest.json


Dananji (approved):
Good catch 🎉

PR #594, opened 2024-07-24
Set accurate duration on progress bar creation

Dananji:
Related issue: #567 

It seems the fixes put in earlier were not working. This PR changes path the duration is carried over to VideoJSProgress component, which seems to be the most stable.


cjcolvar (approved):

Issue #582, opened 2024-07-23
[BUG] Playback head does not progress for audio only item on iOS (iPhone)


elynema:
**Describe the bug**
On iOS iPhone, when an audio only item is played, the playback head does not progress.

**To Reproduce**
Steps to reproduce the behavior, including the results:
1. On an iPhone (iOS 17), go to https://avalon-staging.dlib.indiana.edu/media_objects/rf55z7692. This has been confirmed in Safari, but also reported in other browsers.
2. Click play to start playback.
3. The playback head remains stuck at the beginning, even though the audio is playing.

**Expected behavior**
Playback head should progress as the audio item plays. 

**Screenshots/Screen Recording**

https://github.com/user-attachments/assets/bde06d5e-bc53-46e2-ae11-2a607ac44773

**Environment (please complete the following information):**
 - Device / platform: iPhone
 - OS: iOS 17
 - Browser: Safari (and Chrome?)
- [ ] avalon-dev
- [x ] avalon-staging
- [ ] mco-staging
- [ ] mco-production
- [ ] demo

elynema:
There is a check in Ramp for iOS player put in place for really laggy seeking behavior in native player. iOS devices turns off updating of custom progress bar b/c not necessary in the native player. Because audio player does not use native player, the custom progress bar is not getting updated.

elynema:
@Dananji Tested this in Browserstack today on 2 different iPhones. Although the progress bar shading is advancing, the playhead itself is still not moving. Does this not match what you were seeing?

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBM1FaQmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--733c1ec24dcd6021646f34b89792f16e465fc493/Screen%20Recording%202024-07-29%20at%205.03.57%20PM.mov)

Moving this back to in progress.

elynema:
On avalon-dev, the playback head is now progressing in iPhone iOS17.

However, when I scrub an audio only item, I can hear the weird smooshed audio playing back as I drag the playhead (basically paying at super speed as I scrub). This does not seem to be happening on MCO, but not sure if it 

[...]

PR #595, opened 2024-07-24
Update current time in audio player in iOS context

Dananji:
Related issue: #582

Missed this when working on progress updates issue for audio player in iOS.


cjcolvar (approved):
Good catch!

Issue #591, opened 2024-07-23
[BUG] Scrubbing on iPad (safari) jumps around in video and is slow


elynema:
**Describe the bug**
When scrubbing on an iPad (Safari), the playhead jumps forward, then way backward, then forward again VERY slowly. It makes it look like it's never going to get to where you scrubbed. This happens both when dragging the playhead and when clicking in the progress bar.

I did not notice this behavior as much in the Ramp demo site, so not sure if this is Ramp or Avalon-specific.

**To Reproduce**
1. Go to any record in avalon-staging on an iPad (Safari). Ex: https://avalon-staging.dlib.indiana.edu/media_objects/v405s9361
2. Start playback
3. Try dragging the playhead forward into the video.
4. The playhead jumps forward, then all the way backward, then forward again.
5. Try clicking in the progress bar to scrub forward or backward.
6. The playhead slowly jumps forward, backward, and forward again.

**Expected behavior**
Scrubbing should get to where the user selected within 1-2 seconds and should minimize the frames show in between. In particular, going forward, then backward, makes it look like scrubbing did not work to the user.

**Screenshots**

https://github.com/user-attachments/assets/06b7dab3-2a9c-49b2-a658-e8eb78db35a9

**Smartphone (please complete the following information):**
 - Device: iPad
 - OS: iOS 17
 - Browser: Safari

elynema:
This was the problem that we were originally working on in Safari on desktop. Dananji thinks this is an issue with Safari's timing on player events and believes it is a native video.js issue. If that is the case, there probably is not a reasonable solution.

This should be tested against video.js native player to make sure that is the issue that is happening.

elynema:
Test this in Ramp demo site. If we are happy with how it is working, then we can keep this. If not, revert and continue to work on this for a future release.

Dananji:
The frame lagging when scrubbing is still there in Safari. But it waits for the player to reach the seeked or scrubbed time frame to start playback again. This seems to be the native Video.js behavior when I tested with the Video.js' native progress control component for HLS stream media.

elynema:
@Dananji I just took a quick look at this in iPad on the demo site. I'm not seeing the skipping around, which is an improvement because that is just confusing to users.

I am seeing some fairly significant latency on scrubbing. Sometimes it's 2 seconds and pretty much fine, but sometimes it is 6 seconds or more.

@joncameron Can you also take a look and post if you are seeing the same thing?

joncameron:
[...]

PR #597, opened 2024-07-24
Improve scrubbing in Safari with throttle and debounce on update events

Dananji:
Related issue: #591


src/components/MediaPlayer/VideoJS/components/js/VideoJSProgress.js
```diff
@@ -336,7 +338,6 @@ function ProgressBar({
   const abortableTimeupdateHandler = () => {
     player.on('waiting', () => {
       // Set the player's current time to scrubbed time
```
cjcolvar, line 340: Should this comment get removed too?
```suggestion
```

cjcolvar (approved):
This looks good as far as I can tell but it is hard to know what the full impacts of the change may be.

Issue #590, opened 2024-07-23
[BUG] Clicking in progress bar does not scrub in Android (Chrome)


elynema:
**Describe the bug**
When playing a media item on Android (Chrome), dragging the playhead to scrub works fairly well, but clicking at a point in the progress bar does nothing.

**To Reproduce**
1. Go to any media item in avalon-staging on an Android device (Chrome)
2. Start video playback
3. Try clicking in the progress bar before or after current location
4. Nothing happens

**Expected behavior**
In other contexts, clicking in the progress bar will scrub to the location that was clicked.

**Screenshots**

https://github.com/user-attachments/assets/6ed4f0c5-2a2d-4673-9490-f6256d98bbaf

**Smartphone (please complete the following information):**
 - Device: Android (Pixel was tested)
 - OS: 14
 - Browser: Chrome

joncameron:
👍

Tested on Google Pixel 6 Pro v13.0 in Browserstack.

PR #600, opened 2024-07-25
Use targetTouches in touch event to calculate progress updates

Dananji:
Related issue: #590


cjcolvar (approved):

PR #604, opened 2024-07-30
Align structure highlight marker with hovered progress bar

Dananji:
Fix marker alignment issue introduced in https://github.com/samvera-labs/ramp/pull/587

Before:
Screenshot 2024-07-30 at 1 27 11 PM

After:
Screenshot 2024-07-30 at 1 26 40 PM


cjcolvar (approved):
:+1:

Issue #583, opened 2024-07-23
[BUG] Playback does not begin when audio is muted on iOS (iPhone/iPad)


elynema:
**Describe the bug**
When the audio is muted, playback does not begin for audio only items on iOS (both iPhone and iPad). This was replicated in Safari, but reported to be observed across browsers.

**To Reproduce**
1. Go to https://avalon-staging.dlib.indiana.edu/media_objects/rf55z7692 on iPad or iPhone (iOS17) in Safari.
2. Make sure volume is muted.
3. Start playback.
4. Audio will not start and the playback head will not move.
5. Unmute volume.
6. Playback will start with audio.
7. Note: this can be intermittent. It seems to happen most reliably when the audio is already muted, so after page refresh after you have muted audio.

**Expected behavior**
Playback should begin even if audio is muted.

**Screenshots**

https://github.com/user-attachments/assets/692ef80a-c472-45a5-9195-76b295aad56f

**Smartphone (please complete the following information):**
 - Device: iPad/iPhone
 - OS: iOS17
 - Browser: Safari (and presumably Chrome)

elynema:
Could it possibly be a feature of iOS that playback of an audio item will not begin if it is muted?

Dananji:
I tested the [audio cookbook recipe](https://ramp.avalonmediasystem.org/?iiif-content=https://iiif.io/api/cookbook/recipe/0002-mvm-audio/manifest.json) in this specific environment, and it plays and updates the UI as expected with these settings.

Dananji:
I tested with an [audio manifest](https://dc.library.northwestern.edu/items/774273ef-953d-40a3-8b4e-47ac570e08fd) from one of Northwestern's collections, and observed the same behavior.

I tested this issue in Chrome on iOS in both iPhone and iPad, and observed the same behavior.
In desktop browsers; both Firefox, Chrome played the media when muted but not Safari.

I was not able to find anything that explains this behavior in relation to Apple's system settings on the internet.

One solution we could implement is unmute the audio player on page load in the environments we are seeing this behavior.

elynema:
In Safari, when audio is muted, it never gets to the point where the player is ready to play the media. Same issue when we were trying to activate the buttons in the iOS player b/c the player was not ready enough to have everything ready to play, even when the loaded metadata event has fired. Another event called 'canPlay' - ready state never reaches 4 in Safari when the audio is muted.

abankhele:
[...]

PR #607, opened 2024-07-31
Force the media to load for iOS and Safari on desktop

Dananji:
Related issue: #583

### DIFF
--- a/src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
+++ b/src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
@@ -17,7 +17,7 @@ import {
   useManifestDispatch,
 } from '../../../context/manifest-context';
 import { CANVAS_MESSAGE_TIMEOUT, checkSrcRange, getMediaFragment, playerHotKeys } from '@Services/utility-helpers';
-import { IS_ANDROID, IS_IPAD, IS_MOBILE, IS_SAFARI } from '@Services/browser';
+import { IS_ANDROID, IS_IOS, IS_IPAD, IS_MOBILE, IS_SAFARI } from '@Services/browser';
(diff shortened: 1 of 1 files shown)


src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
```diff
@@ -527,6 +527,20 @@ function VideoJSPlayer({
...
+       * When either player/browser tab is muted Safari and Chrome in iOS doesn't seem to 
+       * load enough data related to audio-only media for the Video.js instance to play 
+       * on page load.
+       * Since, it is not possible to detect muted tabs in JS the condition avoids
+       * checking for muted state altogether.
+       * Without this, Safari will not reach player.readyState() = 4, the state
+       * which indicates the player that enough data is available on the media
+       * for playback.
+       */
+      if (!isVideo && (IS_SAFARI || IS_IOS) && player.readyState() != 4) {
```
cjcolvar, line 541: Are there more ready states than 4?  If so should this be `player.readyState() < 4`?

cjcolvar (commented):
Looks good, but I had one question before approving.

src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
cjcolvar: It looks like ready states only go to 4.
https://docs.videojs.com/player#readyState

cjcolvar (approved):
:+1:

src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
Dananji: Yes, the media would not play unless it reaches 4.

PR #609, opened 2024-08-01
Start from custom manifest start

masaball:
Related issue: #530

### DIFF
--- a/src/components/MediaPlayer/VideoJS/components/js/VideoJSProgress.js
+++ b/src/components/MediaPlayer/VideoJS/components/js/VideoJSProgress.js
@@ -292,11 +292,7 @@ function ProgressBar({
     setTRight(right);
     setTLeft(left);
 
-    setProgress(cTimes.start);
-    setInitTime(cTimes.start);
-
-    setCurrentTime(cTimes.start);
-    player.currentTime(cTimes.start);
+    initializeProgress(cTimes);
 
     /**
      * Using a time interval instead of 'timeupdate event in VideoJS, because Safari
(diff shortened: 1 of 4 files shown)


Dananji (approved):
Looks good 👍 I just have the following questions about refactoring. I don't think they are blockers for merging.

src/components/MediaPlayer/VideoJS/components/js/VideoJSProgress.js
```diff
@@ -440,6 +436,27 @@ function ProgressBar({
...
+      player.currentTime(customStart);
+    } else {
+      setProgress(start);
+      setInitTime(start);
+
+      setCurrentTime(start);
+      player.currentTime(start);
+    }
+  }
+
```
Dananji, line 459: Maybe we could evaluate the condition in the parent function and pass down either `customStart` or `start`to `initializeProgress()` function to setup the initial state without duplicating the setter methods in `initializeProgress` method?

src/services/iiif-parser.js
```diff
@@ -186,6 +186,12 @@ export function getMediaInfo({ manifest, canvasIndex, srcIndex = 0 }) {
     });
     // Set default src to auto
     sources = setDefaultSrc(resources, isMultiSource, srcIndex);
+    // If manifest has a start, set canvas sources' time fragments to match
+    let manifestStart;
+    if (manifest.start !== undefined && manifest.start.source === canvas.id) {
+      manifestStart = manifest.start.selector.t;
+      sources = setDefaultStart(sources, manifestStart, duration);
```
Dananji, line 193: Is it more feasible to read start property this way than using `getCustomStart()` method in the parser?

Dananji (approved):
Looks great 👍

PR #613, opened 2024-08-05
Store captions on/off status in localstorage for session

Dananji:
Related issue: https://github.com/avalonmediasystem/avalon/issues/5651

### DIFF
--- a/src/services/local-storage.js
+++ b/src/services/local-storage.js
@@ -4,8 +4,9 @@ import { useState, useEffect } from "react";
 
 function getValue(key, defaultValue) {
   try {
-    return JSON.parse(localStorage.getItem(key)) || defaultValue;
-  } catch(e) {
+    const storedValue = JSON.parse(localStorage.getItem(key));
+    return storedValue != null ? storedValue : defaultValue;
+  } catch (e) {
     return defaultValue;
   }
 }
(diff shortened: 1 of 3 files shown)


src/services/local-storage.js
```diff
@@ -4,8 +4,9 @@ import { useState, useEffect } from "react";
 
 function getValue(key, defaultValue) {
   try {
-    return JSON.parse(localStorage.getItem(key)) || defaultValue;
-  } catch(e) {
+    const storedValue = JSON.parse(localStorage.getItem(key));
+    return storedValue != null ? storedValue : defaultValue;
+  } catch (e) {
```
Dananji, line 9: Changed this code to check the null value explicitly instead of using `||` operator, since it always evaluated to `true` when `defaultValue` was set to `true`.

cjcolvar: This makes a lot of sense for storing booleans! :100:

cjcolvar (approved):
Looks great!

Dananji:
I read about [nullish coalesing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) just now, and thought that would suit this and would make the code cleaner.

src/services/local-storage.js
```diff
@@ -4,8 +4,8 @@ import { useState, useEffect } from "react";
 
 function getValue(key, defaultValue) {
   try {
-    return JSON.parse(localStorage.getItem(key)) || defaultValue;
-  } catch(e) {
+    return JSON.parse(localStorage.getItem(key)) ?? defaultValue;
```
cjcolvar, line 7: I've never seen this before.  Neat!

cjcolvar (approved):

Issue #596, opened 2024-07-24
Support 200% zoom in media player on mobile devices


elynema:
# Description

For accessibility purposes, it is recommended to support up to 200% zoom (see [WCAG criterion 1.4.4](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-scale.html)). Although this criterion specifically mentions resizability of text, it makes sense that controls also need to be available at 200% zoom.

Using 200% zoom in the browser, we find that controls are lost from Ramp's media player element on mobile devices in both Safari and Chrome.

Potential solutions:

- keep a fixed width for player and require scrolling (not optimal)
- hide player controls behind some kind of widget that can expand 
- wrap player controls to a second row
- don't expand the size of player controls when text is re-sized (not sure if this follow WCAG appropriately) 

# Done Looks Like
- [ ] Test behavior across platforms (Android zoom is different than iOS)
- [ ] Investigate ways to retain elements on screen at appropriate sizes when at high zoom levels
- [ ] Fix things while investigating as needed
- [ ] Player icons are always populated in order (previous button, play/pause button, next button, current time, duration display, volume bar, captions button, track scrubber button, quality selector icon, playback rate button, full-screen toggle) when switching between canvases.

200% zoom in Safari on iPhone 15 (not a mini):

![screenshot-iPhone 15-17 5 (1)](https://github.com/user-attachments/assets/6194a8a6-9289-4675-a69e-0be603c6ec49)

200% zoom (4 notches larger) in Chrome on iPhone 15 (not a mini):
![screenshot-iPhone 15-17 5 (2)](https://github.com/user-attachments/assets/110cd13f-71e8-49c2-9ed1-bd4149aab206)

200% zoom (4 notches larger) in Chrome on Galaxy s24:
![screenshot-Samsung Galaxy S24-14 0 (2)](https://github.com/user-attachments/assets/f7a5ae87-b5ea-452a-9eff-e9fae55a06f0)

elynema:
This problem is potentially less critical on iPhones, as the user only needs to be able to click play, and then controls are accessible from the native player?

elynema:
Youtube does not solve this problem very elegantly in Chrome on iPhone.
[...]

PR #619, opened 2024-08-12
Zoom icons 596

Dananji:
Related issue: #596 

Included in this PR:
- Re-arrange the player controls to provide more space for the controls when the UI is zoomed-in
- Set player's `min-width` so that, all player controls are displayed without overflowing when the UI is zoomed-in

Screenshot 2024-08-12 at 2 27 16 PM

### DIFF
--- a/src/components/MediaPlayer/VideoJS/components/js/VideoJSNextButton.js
+++ b/src/components/MediaPlayer/VideoJS/components/js/VideoJSNextButton.js
@@ -18,6 +18,7 @@ class VideoJSNextButton extends vjsComponent {
   constructor(player, options) {
     super(player, options);
     this.setAttribute('data-testid', 'videojs-next-button');
+    this.addClass('vjs-play-control');
 
     this.mount = this.mount.bind(this);
     this.options = options;
--- a/src/components/MediaPlayer/VideoJS/components/js/VideoJSNextButton.js
+++ b/src/components/MediaPlayer/VideoJS/components/js/VideoJSNextButton.js
@@ -18,7 +18,7 @@ class VideoJSNextButton extends vjsComponent {
   constructor(player, options) {
     super(player, options);
     this.setAttribute('data-testid', 'videojs-next-button');
-    this.addClass('vjs-play-control');
+    this.addClass('vjs-play-control vjs-control vjs-next-btn');
 
     this.mount = this.mount.bind(this);
     this.options = options;
@@ -93,7 +93,8 @@ function NextButton({
   };
 
   return (
-    <div className="vjs-button vjs-control">
+    <React.Fragment>
+      <span className="vjs-control-text" aria-live="polite">Next</span>
       <button className="vjs-button vjs-next-button"
         role="button"
         ref={nextRef}
@@ -103,7 +104,7 @@ function NextButton({
         onKeyDown={handleNextKeyDown}>
         <SectionButtonIcon />
       </button>
-    </div >
+    </React.Fragment >
   );
 }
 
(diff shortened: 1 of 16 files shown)


elynema:
@Dananji Looking at the screenshot, it looks like it's showing the progress bar above the other player controls to create space for the progress bar. I think this looks ok just from the screenshot (although I do notice that the icons on the second row are not vertically aligned).

We'll have to test this out once committed to see how we feel about the min-width.

elynema:
@joncameron You ok with moving the progress bar to a separate line once screen width starts to decrease.

joncameron:
Yep, I'm good with this. @Dananji please go ahead and update the PR from draft so we can merge it in and play around with the change.

Dananji:
I did not test this out in mobile devices as part of the work I did. So, I will test in mobile devices to make sure everything works as expected and make any changes as needed.

Dananji:
Changes made to controls in last commit;

Volume panel change in video player;
Screenshot 2024-08-14 at 2 11 27 PM

Volume panel change in audio player;
Screenshot 2024-08-14 at 2 12 21 PM

src/components/MediaPlayer/VideoJS/components/js/VideoJSNextButton.js
```diff
@@ -18,6 +18,7 @@ class VideoJSNextButton extends vjsComponent {
   constructor(player, options) {
     super(player, options);
     this.setAttribute('data-testid', 'videojs-next-button');
+    this.addClass('vjs-play-control vjs-control vjs-next-btn');
```
cjcolvar, line 21: What's the difference between `vjs-next-btn` and `vjs-next-button`?  Are they supposed to be the same?  (Same question for `vjs-previous-btn` and `vjs-previous-button`.)

cjcolvar (commented):
Looks good overall!  I just had one question.

src/components/MediaPlayer/VideoJS/components/js/VideoJSNextButton.js
Dananji: I think these classes are from an experimental step in the implementation, which was supposed to be removed. Good catch! Let me remove them and re-commit the code.

Dananji:
Yes, I did. And it looked a bit odd with the empty space on top and bottom for me. This wasn't noticeable before with the audio player because the control bar height was smaller then. In this work I increased the control bar height a little bit to make space.
But of course, if the horizontal volume panel with the mute toggle is preferred I can make that change.

src/components/MediaPlayer/VideoJS/videojs-theme.scss
```diff
@@ -0,0 +1,178 @@
+/*
+  Adapted from: https://github.com/videojs/themes/blob/master/city/index.css
+*/
+
+/* Minimum height and width for audio player */
+.vjs-theme-ramp.vjs-audio-only-mode {
+  min-width: 490px;
+  min-height: 50px;
```
cjcolvar, line 8: I'm concerned this may cause problems with existing audio embeds since they only have 40px.

https://github.com/avalonmediasystem/avalon/blob/37cfbea682640796d9d33f4069c8a773537f581e/app/models/concerns/master_file_behavior.rb#L20

Dananji: Yes, it will probably get cut-off. Let me see whether this can be changed.

Dananji: Good catch! 🚀 I changed the `min-height` for audio player to `40px`. It still works with this value as expected.

cjcolvar (approved):
:+1:

Issue #615, opened 2024-08-08
URL Parameters Cause Manifest Request to Fail


joncameron:
# Description
When a media object route is loaded and the URL request has parameters, the manifest is not loaded in Ramp correctly and an error is displayed.

## Examples
https://mco-staging.dlib.indiana.edu/media_objects/9z903m479?_gl=1*56v488*_ga*MTg4MzkzNTc2My4xNzIzMTMwMTQ4*_ga_61CH0D2DQW*MTcyMzEzMDE0OC4xLjEuMTcyMzEzMTQxMi41OS4wLjA.

I can see in network traffic that https://mco-staging.dlib.indiana.edu/media_objects/41688175k?_gl=1*2laqr3*_ga*MTg4MzkzNTc2My4xNzIzMTMwMTQ4*_ga_61CH0D2DQW*MTcyMzEzMDE0OC4xLjEuMTcyMzEzMTQyMS41MC4wLjA/manifest.json is requested and appears to return the media object page HTML document.

Does manifest generation need any parameter information? It could possibly be stripped out before the request for the manifest document.
Is this Ramp needing to strip the parameters before the request, or should Avalon be delivering the manifest document even in this kind of situation?

# Done Looks Like
- [ ] Media object URLs with parameters do not cause errors in manifest generation/retrieval

PR #623, opened 2024-08-15
Sanitize manifest url if params included in middle

masaball:
This commit is to fix an issue found with Ramp embedded in Avalon. If the request url from Avalon has params on the end, Avalon feeds the entire url into Ramp as
`https://example.com/endpoint?params/manifest.json`. This causes media to fail to load because it is not a valid manifest url. By removing the params from the middle of the url, we can get Ramp to request the proper manifest url.

Related issue: #615


Dananji (approved):
Looks good 👍

PR #626, opened 2024-08-16
Reveal player early on progress event with enough data for playback

Dananji:
Reveal player on `progress` event as the browser has enough data to start playback at this point.

I tested this on different browsers on different platforms for an [item from MCO](https://media.dlib.indiana.edu/media_objects/qj72p820d);
- Safari on iPhone: fires `progress` event first, and `loadedmetadata` event fires within ~10ms after that
- Safari on iPads and desktop: fires `loadedmetadata` event first, and the `progress` event fires within less than ~300ms of it
- Chrome on Android: fires `progress` event first, and `loadedmetadata` event fires within less than ~500ms (0.5sec) of it
- Firefox and Chrome on desktop: fires `progress` event first, and `loadedmetadata` event fires within less than ~1000ms (1sec) of it

Since the order in which the events are fired is different on different platforms;
The solution suggested in this PR is, to look for `vjs-disabled` class in the player, which adds the grey overlay on top of the player in both `loadedmetadata` and `progress` events. And whichever event sees it first unveils the player. 
And this should still serve the  purpose of blocking the player until enough data is available for playback.

### DIFF
--- a/src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
+++ b/src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
@@ -468,9 +468,6 @@ function VideoJSPlayer({
 
       player.duration(canvasDurationRef.current);
 
-      // Reveal player once metadata is loaded
-      player.removeClass('vjs-disabled');
-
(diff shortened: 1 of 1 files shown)


cjcolvar (approved):
Great work on this!

Issue #605, opened 2024-07-30
Investigate State Management Alternatives


joncameron:
# Description
What's the current state of the field in state management for React? Investigate and determine which we want to use moving forward as parts of Ramp are refactored. Multiple re-renders are a current paint point in Ramp.

There are multiple options for state management to explore. 
- https://github.com/pmndrs/zustand
- React hooks recently added for rendering strategies

# Done Looks Like
- [ ] Investigation of state management libraries and strategies in React for Ramp use
- [ ] Do we need to swap out for something new, or keep using our existing strategy with updates to hooks and events in React that will resolve rendering issues?

Dananji:
There are multiple state management solutions for React such as; Redux, Jotai, MobX, Recoil, etc. focusing on different areas of expertise and for different scales of applications. And from what I read in blogs, forums, and some documentation it seems Zustand and Recoil are the ones that provide better performance optimization when it comes to state management. 
And out of those 2, Recoil is a fairly new library with a steeper learning curve, suitable for larger applications which require advance state management.

The goal of the investigation was to evaluate whether we need to bring in a new library for state management or whether we could keep using the current solution with some modifications.

So, the comparison was made between Zustand and Context API (existing solution) with custom hooks (new implementation).

## Option 1: Context API with Custom Hooks

Using the existing React Context for state management with additional custom hooks. Custom hooks allow us to abstract reusable logic while the Context API provides a way to share state.

#### Pros:
- Native to React, so no external libraries are added to the project. Less dependencies helps maintain and upgrade things in the future.
- Flexible, since the custom hooks allow us to encapsulate complex logic and reuse it. Helps to keep the components clean and focused on rendering as needed.
- We already have Context setup in Ramp code, therefore less effort than introducing a new library.

#### Cons:
- Performance degradation as Context API can cause unnecessary re-renders in components as global state changes, even if they don't use the specific state that changed.
- Need manual optimization to avoid performance pitfalls, such as [splitting contexts](https://tigerabrodi.blog/optimizing-react-context-and-re-renders) or use of memoization techniques (built-into React 19).
- Customizable but requires additional effort to implement advanced features like asynchronous actions (such as async `React.useEffect`), middleware, or persistent state (such as the custom hook for localStorage), which are provided out of the box in Zustand.

#### Effort:
- Add custom hooks as needed.
- Split contexts to optimize re-renders.

## Option 2: Zustand

[...]

Issue #247, opened 2023-09-22
Expand All/Collapse All for Section Structure


joncameron:
# Description
For first pass implementation, would like ability for a user to be able to expand all sections or collapse all sections using the expand all/collapse button that already exists.

Does Ramp need to provide any additional functionality to enable the expand all / collapse all button in Avalon to work properly? 

# Done Looks Like
- [ ] While a media item is playing, if it is actively playing a section with structure, structure should expand all by default and jump to the actively playing segment.
- [ ] Structure should be expanded by default when media object page loads in Avalon

joncameron:
Re-renders were causing instability in the component, making it difficult to retain the collapsed state. There's a library looked at in coworking time (https://github.com/pmndrs/zustand) that is for subscribing to state events and could be helpful.

This will probably require a refactor to the existing state management in Ramp.

elynema:
From Chris:

Here's what I found with a solr query and computations on the ruby side:
0 number of sections: 500
1 number of sections: 88603                                                                                               
2 number of sections: 69750                                                                                               
3 number of sections: 7409                                                                                                
4 number of sections: 6497                                                                                                
5 number of sections: 916                                                                                                 
6 number of sections: 2805                                                                                                
7 number of sections: 253                                                                                                 
8 number of sections: 854                                                                                                 
9 number of sections: 116                                                                                                 
10 number of sections: 339                                                                                                
11 number of sections: 70                                                                                                 
12 number of sections: 195                                                                                                
13 number of sections: 63                                                                                                 
14 number of sections: 83                                                                                                 
15 number of sections: 51
16 number of sections: 83
17 number of sections: 28
18 number of sections: 49
19 number of sections: 21
20 number of sections: 51
21 number of sections: 17
22 number of sections: 23
23 number of sections: 13
24 number of sections: 54
25 number of sections: 12
26 number of sections: 27
27 number of sections: 11
28 number of sections: 11
29 number of sections: 15
30 number of sections: 4
31 number of sections: 8
32 number of sections: 11
33 number of sections: 6
34 number of sections: 10
36 number of sections: 6
37 number of sections: 6
39 number of sections: 3
41 number of sections: 2
43 number of sections: 4
45 number of sections: 4
47 number of sections: 3
48 number of sections: 5
49 number of sections: 1
52 number of sections: 3
53 number of sections: 1
55 number of sections: 8
56 number of sections: 3
58 number of sections: 1
60 number of sections: 1
61 number of sections: 1
[...]

PR #627, opened 2024-08-26
Initial attempt at retaining collapse/expand sections with custom hooks

Dananji:
Related issue: #605 

This work is done as PoC to evaluate the eligibility of custom hooks to solve the re-rendering issues we were having with structured navigation collapse/expand feature (https://github.com/samvera-labs/ramp/issues/247).

If we decide to go forward with state management with React context API with custom hooks as a result of the investigation (#605), this work can be re-purposed as a starting point.

https://github.com/user-attachments/assets/721ec64f-4588-4260-ace2-9117177b1c73


Dananji:
Closing in favor of #643

Issue #624, opened 2024-08-16
Player replays last section/single section item in a loop


Dananji:
### Description

For single section items the player is replaying the section.  For multi-section items it is replaying the last section.

### Done Looks Like
- [ ] Once player reaches the end of a single section item
  - [ ] stop playback
  - [ ] don't reset current time to zero and show replay icon instead of play/pause icon in the controls

- [ ] Once player reaches the end of the last section in multi-section item  or a playlist
  - [ ] stop playback
  - [ ] stay on last section / last playlist item (don't go back to the beginning)
  - [ ] don't reset current time to zero and show replay icon instead of play/pause icon in the controls
  - [ ] replay icon replays the current section / playlist item

elynema:
Hold this until Ramp state management / code cleanup is complete.

joncameron:
Worked as expected in Ramp demo site for single- and multi-section items.

PR #628, opened 2024-08-27
Fix player replaying last section/single section item in a loop

Dananji:
Related issue: #624


cjcolvar (commented):
Looks good as far as a I can tell.

PR #629, opened 2024-08-27
Limit time tooltip display offset to playable range width

Dananji:
When mouse pointer is hovering over the blocked parts of the progress-bar, the time tool-tip is displayed outside of the progress-bar container.
This PR fixes the alignment of the time tool-tip by limiting the calculated offset to always align tool-tip with the playable range of the media.

Before:
Screenshot 2024-08-27 at 4 03 51 PM

After:

Screenshot 2024-08-27 at 4 05 02 PM


cjcolvar (approved):
Looks tricky!  I don't totally get it but looks fine to me.

PR #642, opened 2024-09-17
Make `withCredentials` work with videojs 7.x and 8.x

mbklein:
Instead of using the `hlsOptions` property to pass the credentials flag down to VideoJS, set `withCredentials: true` on each of the sources.


cjcolvar (approved):
Looks good to me.  Thanks for your hard work on this!

PR #643, opened 2024-09-18
State management refactor for structure

Dananji:
Related issue: #630 

Use custom hooks for structured navigation for state updates and subscriptions limiting re-renders of the component tree in the `StructuredNavigation` component. This lets us to use collapsible sections and then keep it collapsed or expanded while the state is getting updated in the background.

For Avalon media object manifest: https://avalon-dev.dlib.indiana.edu/media_objects/g158bh28p/
Screenshot 2024-09-18 at 2 18 23 PM

For cookbook recipe: https://iiif.io/api/cookbook/recipe/0065-opera-multiple-canvases
Screenshot 2024-09-18 at 2 19 08 PM

For cookbook recipe: https://iiif.io/api/cookbook/recipe/0026-toc-opera
Screenshot 2024-09-18 at 2 18 49 PM


cjcolvar (approved):
This looks really good!

Not in scope of this PR but when looking through it I thought the cyclical dependence of SectionHeading on List and List on SectionHeading is confusing.  I wonder if having an abstract component with subclasses might help scope and conceptualize the different uses of the two components.

Dananji:
> Not in scope of this PR but when looking through it I thought the cyclical dependence of SectionHeading on List and List on SectionHeading is confusing. I wonder if having an abstract component with subclasses might help scope and conceptualize the different uses of the two components.

That sounds good. I created a new issue to work on this https://github.com/samvera-labs/ramp/issues/648

PR #647, opened 2024-09-19
Custom hook for MediaPlayer and VideoJSPlayer handling

Dananji:
Related issue: #630 

Includes custom hook implementations for player related state updates.

For some of the functionality in the custom hooks, unit tests are added. Most of the code in these hooks are covered by the unit tests in respective components' test-cases. Test coverage in the new module `@Services/ramp-hooks.js` added in this work (#630) is around 70%. The parts that are not covered in tests are either impossible/hard to write tests with just the custom hook functionality.

### DIFF
--- a/src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
+++ b/src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
@@ -1,4 +1,4 @@
-import React, { useMemo } from 'react';
+import React, { useMemo, useRef } from 'react';
 import PropTypes from 'prop-types';
 import videojs from 'video.js';
 import throttle from 'lodash/throttle';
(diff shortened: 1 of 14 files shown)


src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
```diff
@@ -56,293 +52,171 @@ function VideoJSPlayer({
...
+  const { messageTime } = useShowInaccessibleMessage({ lastCanvasIndex });
 
-  let cIndexRef = React.useRef();
-  cIndexRef.current = canvasIndex;
-  const setCIndex = (i) => {
-    _setCIndex(i);
-    cIndexRef.current = i;
-  };
+  const canvasIsEmptyRef = useRef();
+  canvasIsEmptyRef.current = useMemo(() => { return canvasIsEmpty; }, [canvasIsEmpty]);
```
cjcolvar, line 80: This is a neat pattern for updating a ref using hooks.  How does this compare with what react would do with useState and not a ref?

cjcolvar (approved):

src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
Dananji: There are couple of advantages to using refs in place of state;
- Using refs in place of state helps to control re-renders of the component as state variables automatically triggers a component re-render
- In this particular instance, when using these variables inside player's event handlers it seems variables defined as refs carry the most recently updated value rather than the variables defined as state.
- Using `useMemo` to update the refs ensures that the value of the variable is persisted across re-renders without corruption, since it is only updated when the dependency of `useMemo` is changed.

PR #652, opened 2024-09-27
Custom hook for state management in Transcript component

Dananji:
Related issue: #630 

This work moves state management in Transcript component to a custom hook.
Since `Transcript` component is built as an independent component from global state management using context providers it doesn't have the issue other components have with unnecessary re-renders.
But it can be used within context providers, without breaking its independent nature.
This refactor helps up to make use of the global state IF the component is wrapped inside the context providers.
One advantage of this is, `Transcript` component doesn't need to spin up a unnecessary JS interval to read the player instance from the page when the player is available via context providers, while still supporting its independent implementation.

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -86,7 +86,6 @@ const App = ({ manifestURL }) => {
   );
 };
 
-
 /*Reference: https://accessible-react.eevis.codes/components/tabs */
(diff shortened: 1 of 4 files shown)


cjcolvar (approved):
Looks great!  I really like how this makes the Transcript component much smaller!

PR #662, opened 2024-10-10
Fix minor bugs in player introduced in refactor/cleanup work

Dananji:
This PR includes the following;
1. Cleanup of `playerFocusElement` global state variable, as this is not needed anymore to retain focus for previous/next button keyboard navigation with the changes done on these custom components with new implementation
2. Remove `float: inline-end` styling with `display: inline-block` for player controls to right align some of the controls in the player, as this reverses the tab order of the controls when using keyboard navigation. Replace this styling to use `display: flex`
3. Fix a bug in the state updates for persisting the player's play/pause status as `isPlaying` in global state with hotkey implementation and pointer events
4. Set aspect ratio for inaccessible message display as 16/9 (equivalent to video player) when switching from an audio item to inaccessible item
5. Attach player hotkeys event handler upon player creation, as hotkeys doesn't work as expected with the previous implementation when the first item is an inaccessible item.

### DIFF
--- a/src/services/ramp-hooks.js
+++ b/src/services/ramp-hooks.js
@@ -330,27 +330,6 @@ export const useVideoJSPlayer = ({
   };
 
   useEffect(() => {
-    /*
-      This event handler helps to execute hotkeys functions related to 'keydown' events
-      before any user interactions with the player or when focused on other non-input 
-      elements on the page
-    */
-    document.addEventListener('keydown', (event) => {
-      const result = playerHotKeys(event, playerRef.current, canvasIsEmpty);
-      // Update player status in global state
-      switch (result) {
-        case HOTKEY_ACTION_OUTPUT.pause:
-          handlePause(false);
-          break;
-        case HOTKEY_ACTION_OUTPUT.play:
-          handlePause(true);
-          break;
-        // Handle other cases as needed for each action
-        default:
-          break;
-      }
-    });
-
     // Dispose Video.js instance when VideoJSPlayer component is removed
     return () => {
       if (playerRef.current) {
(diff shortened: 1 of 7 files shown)


src/services/ramp-hooks.js
```diff
@@ -386,12 +388,12 @@ export const useVideoJSPlayer = ({
 
       // Update player status in state only when pause is initiate by the user
       player.controlBar.getChild('PlayToggle').on('pointerdown', () => {
-        handlePause();
+        handlePause(player.paused());
```
cjcolvar, line 391: I'm a bit confused.  Should this be `player.paused()` or `!player.paused()`?  The `handlePause` function calls this argument `isPlaying` which seems like it would be the opposite of `player.paused()`.

Dananji: Yes, this feels wrong at a first glance 🙂 Because we expect the player status to be changed immediately upon the user interaction, but it seems these event handlers capture the player status before it gets altered with the respective action. And for that reason this works as expected by just using `player.paused()`.

For an example;
1. Using `PlayToggle` button:
  - Initially `player.paused()` = `true`
  - Click `PlayToggle` button => `player.paused()` = `true` from initial value before this action alters the player's status; `isPlaying` = `true`
  - Click `PlayToggle` button again => `player.paused()` = `false`; `isPlaying` = `false`

2. Using pointer on top of video:
  - Initially `player.paused()` = `true`
  - Click on top of video player => this event is captured by `BigPlayButton` overlay, therefore not captured in these event handlers. But this is captured in the `play` event handler in `VideoJSPlayer` module and it sets `isPlaying` = `true`
  - Click on top of video player again => `player.paused()` = `false`; `isPlaying` = `false`

I tested this again with logs, to make sure it works as expected for both scenarios where there are inaccessible items in the Manifest and all playable items in the Manifest.

cjcolvar (approved):
Looks good!

Issue #637, opened 2024-09-13
Refine custom progress bar for smooth scrubbing in all platforms


Dananji:
### Description

A considerable amount of change was made to the custom progress bar component as it was re-architected to use Video.js' `SeekBar` component instead of building the HTML as part of #618.
With this work the scrubbing in iOS is not as smooth as it is in the native VideoJS `SeekBar` component (the expected outcome).
The hierarchy of component structure for `ProgressControl` in Video.js is as follows;
```
├─┬ ProgressControl
│ │ └─┬ SeekBar
│ │   ├── LoadProgressBar
│ │   ├── MouseTimeDisplay
│ │   └── PlayProgressBar
``` 
And in the current implementation we are using `SeekBar` and its children, not `ProgressControl`.
I tried attaching the custom seek bar (the implementation merged in #635) to `ProgressControl` instead of exporting as it is with the following code;
```
const ProgressControl = videojs.getComponent('ProgressControl');
class VideoJSProgress extends ProgressControl {
  constructor(player, options) {
    super(player, options);
    this.addClass('vjs-custom-progress-bar');
    this.removeChild('SeekBar');
    this.addChild('CustomSeekBar');
  }
}

videojs.registerComponent('CustomSeekBar', CustomSeekBar);
videojs.registerComponent('VideoJSProgress', VideoJSProgress);

```
 But the child component in `MouseTimeDisplay` was facing issues with finding the custom seek bar fed into `ProgressControl`. To solve this we can try customizing the children components that are having issues?

### Done Looks Like
- [ ] Attach custom seek bar to `ProgressControl` 
- [ ] Fix issues with the children components with the approach
- [ ] Test in all supported browser and platform combinations

joncameron:
Some is changed in recent refactor work but not all. Mobile (iOS) scrubbing is still off and there's some more work to build on the native VideoJS component to update the behavior.

elynema:
Dananji noticed some jumping in seeking in Browserstack, but it was smoother on her device. Android is smoother than iOS.

elynema:
I'm going to go ahead and close this as we've had lots of testing of scrubbing across all platforms and opened multiple other tickets.

PR #664, opened 2024-10-14
Refine progress-bar by using VideoJS' native ProgressControl

Dananji:
Related issue: #637

### DIFF
--- a/src/components/MediaPlayer/MediaPlayer.js
+++ b/src/components/MediaPlayer/MediaPlayer.js
@@ -156,12 +156,7 @@ const MediaPlayer = ({
             'fullscreenToggle',
             // 'vjsYo',             custom component
           ],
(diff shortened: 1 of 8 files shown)


cjcolvar (approved):
Looks good as far as I can tell.  It would be nice if there was a cleaner way to override the seek bar implementation in VideoJS so you don't have to override the progress control suppress the native seekbar and forward events but that is probably unavoidable.  Overall still seems better than everything custom as it was before.

Dananji:
Yes, this is not ideal 💯  There's probably a better way to do this, which I want to explore as I get time in the future 🙂 
I tried removing the native `SeekBar` and attaching the custom implementation and it breaks the other child components. So, I tried overriding the code that breaks in the child components (especially in `TimeTooltip` component) which didn't seem to work.

Issue #657, opened 2024-10-08
Replace new volume control


elynema:
# Description

The new volume control button in Ramp that is shaped like a triangle does not allow for as easy adjustment of volume as the slider and does not provide a single click to mute/ un-mute.

Consider bringing back the vertical slider. We have numerous other widgets in Ramp that overlay the progress bar with a menu, so it seems like it could work for the vertical slider as well. Alternatively, all the popups that are used by widgets, including the vertical slider, could be placed above the progress bar.

# Done Looks Like
- [x] Volume control component changed to the previous style: click to enable/disable with a slider overlay for the volume

Dananji:
Note for testing:
- All popup menus were moved to display above progress-control
- All popup menus (captions, quality selection, file download, playback rate) are now displayed only on `click`/`keydown` (`Enter`/`Space` on focus, when using keyboard navigation) events. Previously these menus were shown on both `click`/`keydown`/`hover` event.

elynema:
Tested across Android, iPad, Safari and Chrome desktop, as well as audio and video embeds. The audio control itself looks like it is working fine in al these contexts.

PR #669, opened 2024-10-17
Replace volume control with muteToggle/volumePanel, popup menus above progress-bar

Dananji:
Related issue: #657 

This PR includes the following changes;
1. Revert volume bar to the previous state where video players have vertical volume control and audio players have sticky horizontal volume control as follows.

Screenshot 2024-10-17 at 4 54 48 PM
Screenshot 2024-10-17 at 4 55 30 PM

2. Move all popup menus above the progress-control. As part of this work display of the popup menus on `hover` event on the buttons are removed. This seems to be the common practice in other platforms e.g. LinkedInLearning and YouTube.

### DIFF
--- a/src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
+++ b/src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
@@ -252,6 +252,8 @@ function VideoJSPlayer({
     /*
       Update player control bar for;
        - track scrubber button
+       - volume panel
+       - if tracks exists: captions button for video players
        - appearance of the player: big play button and aspect ratio of the player 
         based on media type
        - file download menu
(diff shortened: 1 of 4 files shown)


src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
```diff
@@ -277,6 +279,27 @@ function VideoJSPlayer({
...
+       * For mobile devices the player uses MuteToggle for both audio
+       * and video.
+       */
+      if (!IS_MOBILE) {
+        controlBar.removeChild('volumePanel');
+        if (!isVideo) {
+          controlBar.addChild('volumePanel', { inline: true }, volumeIndex);
+        } else {
+          controlBar.addChild('volumePanel', { inline: false }, volumeIndex);
+        }
```
cjcolvar, line 295: Does this duplicate the change in MediaPlayer.js or is it needed to place the volume control in the proper order of controls?  If the later, could this be simplified to match the config in MediaPlayer.js?
```suggestion
        controlBar.addChild('volumePanel', { inline: !isVideo }, volumeIndex);
```

src/components/MediaPlayer/VideoJS/videojs-theme.scss
```diff
@@ -52,135 +62,61 @@
...
-  height: 60%;
-  margin-top: 0.5em;
-}
-
-/* Volume stuff */
-.vjs-theme-ramp .vjs-volume-panel:hover .vjs-volume-control.vjs-volume-horizontal {
-  height: 100%;
+.vjs-theme-ramp .vjs-menu li.vjs-menu-title {
+  text-transform: capitalize;
 }
```
cjcolvar, line 109: I'm thinking that we need to keep hover display of the volume popup since click toggles mute or am I not understanding this correctly?

src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
Dananji: It's the later.

src/components/MediaPlayer/VideoJS/videojs-theme.scss
Dananji: Hover display still works for the vertical volume panel. The removed CSS here was for the triangle volume bar in the previous implementation.

cjcolvar (approved):

Issue #645, opened 2024-09-18
Implement Expand/Close button into StructuredNav Component


Dananji:
### Description

With the on-going state management refactor using custom hooks (#630), we are able to implement expand/close functionality for each section in the structured navigation in Ramp.
We were not able to do this before due re-rendering in StructuredNavigation component, therefore this functionality was implemented in Avalon code.

### Done Looks Like
- [ ] Implement expand/close sections button in StructuredNavigation component
- [ ] Use a prop that goes into StructuredNavigation component to turn on/off the display of this button
- [ ] Document props in styleguidist docs and GitHub wiki
- [ ] Test functionality in all supported platforms
- [ ] Removed related code in Avalon when this is brought into Avalon

### Related work
https://github.com/samvera-labs/ramp/pull/644

elynema:
In my testing, across all platforms I have observed that the button text does not update if all sections are manually opened or closed:

1. Click button to close all sections
2. Expand all sections individually
3. Button still says 'Expand [x] sections"
4. Click button to expand all sections
5. Manually close all sections individually.
6. Button still says "Close [x] sections" even though all sections are closed.

This happens even if there is only 1 section.
 
Also across all platforms, if the user loads page and clicks "Close [x] sections" button, then starts playback, the first section does not pop open when playback starts (the second one does).


[Uploading... Screen Recording 2024-11-08 at 1.21.16 PM.mov]

elynema:
Note that aside from these minor issues, the new button is working wonderfully!! I asked Dananji to take a look at the last issue reported here (the first section does not open when playback begins if all sections are closed prior to playback). 

@joncameron Do you think we should create new issues for either of the other minor issues above?

joncameron:
I created issues for both of these: #731 and #732.

joncameron:
Tested and closed #731 and #732; closing this issue as well. Working great!

PR #671, opened 2024-10-22
Collapse/expand sections button

Dananji:
Related issue: #645 

https://github.com/user-attachments/assets/0ad0a41a-9e01-491c-89bb-a430be5c8667

When `structures` property has a root Range, it is rendered as a non-collapsible section as it contains all of the structure in the Manifest. e.g. from [cookbook toc recipe](https://iiif.io/api/cookbook/recipe/0065-opera-multiple-canvases);
cookbook-example-root


src/components/StructuredNavigation/NavUtils/SectionHeading.js
```diff
@@ -30,12 +30,13 @@ const SectionHeading = ({
   sectionRef,
   structureContainerRef,
 }) => {
-  // Always collapse root structure element
-  const [isOpen, setIsOpen] = useState(isRoot);
+  const { isCollapsed } = useCollapseExpandAll();
+  // root structure items are always expanded
+  const [sectionIsCollapsed, setSectionIsCollapsed] = useState(
+    isRoot ? false : true || isCollapsed);
```
cjcolvar, line 36: Is this always going to be `true` for non-root structure items?

src/components/StructuredNavigation/NavUtils/SectionHeading.js
```diff
@@ -45,9 +46,17 @@ const SectionHeading = ({
...
-    setIsOpen
+    setSectionIsCollapsed
   });
 
+  // Collapse/Expand section when all sections are collapsed/expanded respectively
+  useEffect(() => {
+    // Do nothing for root structure items
+    if (!isRoot) {
+      isCollapsed ? setSectionIsCollapsed(true) : setSectionIsCollapsed(false);
+    }
```
cjcolvar, line 57: How about simplifying this?
```suggestion
    if (!isRoot) {
      setSectionIsCollapsed(isCollapsed);
    }
```

src/services/ramp-hooks.js
```diff
@@ -652,7 +659,7 @@ export const useActiveStructure = ({
     // Do not mark root range as active
     if (isCurrentSection && !isRoot) {
       // Collapse the section in structured navigation
-      setIsOpen(true);
+      setSectionIsCollapsed(false);
```
cjcolvar, line 662: This comment is confusing given the next line appears to mark the section as not collapsed.  Which is correct?

cjcolvar (commented):
Looks really good overall!

src/components/StructuredNavigation/NavUtils/SectionHeading.js
Dananji: This sets the value on initial render, and then changed as user interacts with the structure and the collapse/expand all sections button.

src/services/ramp-hooks.js
Dananji: True. I'll change the comment. It should be section is expanded by setting isCollapsed to false.

src/components/StructuredNavigation/NavUtils/SectionHeading.js
cjcolvar: I think I'm confused about the boolean logic and what it trying to be achieved with `|| isCollapsed`.  If I'm reading it correctly `isCollapsed` isn't being used at all because `true` will always be returned when `isRoot == false`.

Dananji: Correct! Sorry, I was mistaking this with a previous iteration of the implementation I had in mind. I'll fix it. Thanks!

cjcolvar (approved):
:+1:

Issue #676, opened 2024-10-23
[BUG] media player is not re-sizing to width of browser


elynema:
# Description

For both audio and video items, the media player is not re-sizing to the width of the browser on the playlist page. This requires the user to scroll the browser horizontally in order to access all the widgets. So far, this has been observed in Safari on iPhone as well as Chrome on desktop (when the browser is resized down). This is happening even when the browser is not zoomed at all and is happening for both audio and video on both playlist and media item pages.

![screenshot-iPhone 15 Pro-17 0 (3)](https://github.com/user-attachments/assets/b610abef-e923-4cfc-91b2-0e9ac2114504)

# Done Looks Like
- [ ] Media player resizes to fit the entire width, including components, to within the browser viewport width
- [ ] Test zoom levels and investigate resizing strategy to determine where to make changes
- [ ] There is enough space to accommodate the popup menu for the playback speed widget.

Dananji:
I was not able to reproduce this issue for Chrome on both my machine and BrowserStack for desktop.

elynema:
Tested this morning and I also was not able to reproduce this specific issue in Chrome or Safari today. However, I am seeing an overlap issue in what is probably tablet mode in both Chrome and Safari. 

![image (10)](https://github.com/user-attachments/assets/c3d081ed-6770-47e1-829e-ce251a971f1c)

elynema:
@Dananji Chrome and Safari on desktop is working fine for me now for both issues reported above, but the video player is still overflowing the page horizontally in Safari on iPhone in Browserstack.

![screenshot-iPhone 15-17 0 (8)](https://github.com/user-attachments/assets/e9a64d5d-7ce2-43bb-a311-556ab4f5e0ab)

Dananji:
@elynema Latest changes from the Ramp build were not deployed properly in `avalon-dev`, so it was still on the older CSS. Chris re-compiled them, and now latest CSS changes are in place in `avalon-dev`.

elynema:
This is looking good on iPhone, but Android is still overflowing. I see you have #696 that isn't merged yet - is that a resolution to this issue?

![screenshot-Samsung Galaxy S24-14 0 (4)](https://github.com/user-attachments/assets/866dda07-0e1d-4346-a169-42370975aadf)

elynema:
Tested and working fine on Chrome and Safari on desktop

@Dananji Two issues:

[...]

PR #684, opened 2024-10-25
Change min-width and min-height of the player for iOS

Dananji:
Related issue: https://github.com/samvera-labs/ramp/issues/676


cjcolvar (approved):

PR #685, opened 2024-10-25
Regression bug: display error for empty Manifest w/o crashing

Dananji:
Display specific error message without crashing Ramp when an empty Manifest is provided. This is a regression introduced in the refactor/upgrade work after 3.2.1 release.


cjcolvar (approved):
Looks good to me.

Issue #673, opened 2024-10-22
[BUG] Problems navigating playlist items when first item is inaccessible


elynema:
# Description

If the first item in a playlist is inaccessible, and the user clicks Next (before auto-advance) or selects the next playable item in the playlist items, the next item to play will loop rather than advancing when it completes and all the other playlist items will not be selectable (nothing happens when you click on them).

If the user allows the first inaccessible item to advance without clicking Next or selecting another item, it will auto-advance after 10 seconds, skipping the next item in the playlist.

This was observed in Safari and Chrome on desktop, as well as Safari on iPhone.

Example playlist: https://avalon-dev.dlib.indiana.edu/playlists/5 

Auto-advance skips second item in playlist:

https://github.com/user-attachments/assets/4f564d74-fc46-4790-b9ff-790ec2c71bcd

Manually advancing the inaccessible item causes the next item to loop and then nothing happens when selecting other playlist items:

https://github.com/user-attachments/assets/68fc732c-c044-4eae-a39c-1e08632ec7e8

elynema:
Note: I am also seeing this now for inaccessible items that are not the first item. If autoadvance is used, playback will skip the item immediately following the inaccessible item. Allow this playlist to autoadvance while logged out: https://avalon-dev.dlib.indiana.edu/playlists/178. After the second inaccessible item's timer runs out, the highlighting skips to the fourth item and the inaccessible error displays, but the audio from the skipped third item begins to play.

Probably related to ramp #680.

Dananji:
> If the first item in a playlist is inaccessible, and the user clicks Next (before auto-advance) or selects the next playable item in the playlist items, the next item to play will loop rather than advancing when it completes and all the other playlist items will not be selectable (nothing happens when you click on them).

I was not able to reproduce this in `avalon-dev` in both Chrome and Safari on desktop.

> If the user allows the first inaccessible item to advance without clicking Next or selecting another item, it will auto-advance after 10 seconds, skipping the next item in the playlist.

I was able to reproduce this in `avalon-dev` in Chrome but not in the demo-site (for the same playlist).

elynema:
Re-tested and agree wtih Dananji that I'm no longer seeing issues when using Next button or playlist items to navigate past the first inaccessible item in avalon-dev on Chrome, Safari, or iOS. Still seeing issues with the auto-advance as the new Ramp build isn't brought into avalon-dev yet.

[...]

PR #690, opened 2024-10-28
Create inaccessible message interval only in a single custom hook

Dananji:
Related issue: #673 

Only invoke `createDisplayTimeInterval()` method within `useShowInaccessibleMessage` custom hook. When this was called within `initCanvas()` in `useSetupPlayer` hook, it creates multiple intervals disabling the remaining time updates in the timer on display.
Use `canvasIndex` instead of `canvasIndexRef` in the dispatch to update manifest-state when timer ends. 
It was observed that, in Safari and Chrome using `canvasIndexRef` causes to skip an item when auto-advancing (most probably due to timing of these updates are getting overlapped) in avalon's playlists where there are inaccessible items. However this was not seen in these browsers in the demo site.

### DIFF
--- a/src/services/ramp-hooks.js
+++ b/src/services/ramp-hooks.js
@@ -107,7 +107,6 @@ export const useSetupPlayer = ({
   const manifestState = useContext(ManifestStateContext);
   const {
     allCanvases,
-    autoAdvance,
     canvasIndex,
     customStart,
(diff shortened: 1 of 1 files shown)


cjcolvar (approved):
Great work!  This seems like it was really tricky to find!

Issue #622, opened 2024-08-15
Track scrubber missing aria-valuenow attribute


elynema:
# Description

The SiteImprove accessibility checker browser plugin is indicating that there is an element with role="slider" that is missing the aria-valuenow attribute on the media object page. Looking at where that role occurs in the code, I believe that is in the Track scrubber. It does not appear to have an aria-valuenow attribute even when expanded.

I'm not seeing this issue show up in the SiteImprove site scanner or in WAVE, so it could be a false positive?

![image](https://github.com/user-attachments/assets/899de095-afe9-4ba3-9399-4ad0e1cf2ca4)

# Done Looks Like
- [ ] Evaluate whether this is someone requiring a change or fix
- [ ] The missing value does appear in other places in Avalon; we can look to that for other examples

elynema:
Not seeing the error for the track scrubber missing an ARIA attribute anymore, so that seems to be resolved.

@Dananji When I ran the SiteImprove tool against this page: https://avalon-dev.dlib.indiana.edu/media_objects/r207tp32d, I'm getting 2 errors that occur multiple times. One is about spacing around buttons / links, which I think is WCAG 2.2. The other is about hidden elements having focusable content. That sounds like part of WCAG 2.1.

![image](https://github.com/user-attachments/assets/9f2dd3cd-81d4-44ee-8f69-71bb5a1ea755)

Are you seeing this in your browser? I did try navigating the page via keyboard, and it didn't seem like the cursor was navigating to anything hidden.

Dananji:
@elynema Yes, I'm seeing these only in SiteImprove.
And I tested another item that is public on both `avalon-dev` and on Ramp demo-site, the errors are only reported in `avalon-dev`. And the elements it's complaining about are the VideoJS error display modal and the captions setting modal, which are native to VideoJS.

elynema:
@Dananji It makes sense why the spacing issues might not show up on the demo site, as the player is larger in that context. But I don't understand how the focus errors that refer to video.js components can show up in avalon-dev but not in Ramp.

Is there a difference in version of React or version of video.js or Bootstrap styling that could somehow cause the difference?

PR #691, opened 2024-10-29
Fix a11y issues found with Siteimprove, WAVE, and Axe

Dananji:
Related issue: #622 and other a11y issues in components


cjcolvar (approved):

PR #692, opened 2024-10-29
Fix sections UI for non-collapsible structures

Dananji:
This includes the following fix for non-collapsible structures when sections portion of the `StructuredNavigation` component is turned on. This mimics the previous UI for sections in Avalon for non-collapsible structures.

Before:
Screenshot 2024-10-29 at 1 31 11 PM

After:
Screenshot 2024-10-29 at 1 30 47 PM


cjcolvar (approved):
:+1:

Issue #677, opened 2024-10-23
[BUG] Can't scrub before beginning playback for media item (iPhone, Safari)


elynema:
# Description

On the media item page, if you click in the progress bar / move the playhead to scrub before beginning playback, when you hit play, the item begins playing at the beginning. This may be an interaction between Ramp and native iOS player?

In addition, once the native player opens and playback begins, scrubbing does not work at all in the native player. You can adjust the progress, but the video just keeps playing from the beginning.

Note: If the user attempts to move the playhead before beginning playback, the player spins for a long time in Safari before the playback begins at the beginning of the media item in the native player. If the native player is subsequently closed, the media player is still spinning on the media item page.

# Done Looks Like
- [ ] Playback begins at playhead position, even when the playhead is adjusted before initial playback
- [x] Scrubbing is possible in the native player, when when the playhead is adjusted before playback.
- [x] Users can scrub either by moving the playhead or clicking in the progress bar and the media player does not get stuck (moving to native player and back results in endless indication of buffering and media playback doesn't resume)

joncameron:
Setting the source to be offset has been a previous way to mark to the native player that playback should begin at a certain point. In Safari on iOS the event that sets this value isn't going through and playback always begins the start.

elynema:
@Dananji This is working some / most of the time on iPhone. Two issues I have run into:

1. After the scrub, the thumbnail doesn't change in the media player. So if you've scrubbed to minute 4:00 of lunchroom manners, you're still seeing the initial thumbnail that shows the title slide. I don't think that's a blocking issue, but wanted to mention in case it is fixable. If you start and pause playback and then scrub, the thumbnail updates.

Thumbnail does not update:

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK3lqQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b962a28a4857df1f01292ea69f7b91c11215069d/Screen%20Recording%202024-11-07%20at%2012.01.13%E2%80%AFPM.mov)



2. Sometimes the first time you click in the progress bar to scrub, the progress bar jumps forward and then back to the beginning. It generally works on the second click. I was not able to get this to replicate while making a video, so I don't have a recording. I'd be interested to see if you can get it to replicate. This is also maybe not a blocker as it's usually only a problem on the first click.

[...]

PR #708, opened 2024-11-01
Update progress-bar CSS on fullscreen exit in iOS

Dananji:
Related issue: #677 

Partial solution for this issue in #707.

This PR updates the progress-bar's played range CSS in the VideoJS player to reflect the time updates made in the native player with seek/scrub actions, when the user exits the native iOS player.


cjcolvar (approved):
:+1:

Issue #687, opened 2024-10-28
Add Padding to Progress Bar


joncameron:
# Description
The progress bar currently extends to the full width of the media player. Padding around the progress bar would provide easier scrubbing around the beginning and ending around recordings: it's difficult to click and interact with playhead when at 00:00 (and some of the playhead is not visible).

# Done Looks Like
- [x] Center align the progress bar (audio player) the top of the control area bar and the player controls
- [x] 12 pixels of margin or padding to the left and right of the progress bar
- [x] Text tooltip for current time can continue floating above the player
- [x] Keep height of the media player the same
- [x] This design needs to be tested as well with the height and width of Avalon embeds in the wild

# Mockup
![Screenshot 2024-10-30 at 12 28 38 PM](https://github.com/user-attachments/assets/d216891c-d657-47d0-8c45-bdce4f414cb1)

# Current audio player
![Screenshot 2024-10-30 at 3 13 09 PM](https://github.com/user-attachments/assets/35d7e1b9-ad40-4b5a-bdcc-cff006b296e6)

elynema:
I think the new design is looking much improved! Seeing the left/right padding, new size of progress bar, and vertical centering on all devices. No animation when you hover over the progress bar.

I have noticed one issue. When I scrub by clicking on the progress bar or dragging the playhead, the progress bar gets a sort of yellow glow around it. This remains until I pause and restart the audio/video or click elsewhere on the page. This doesn't happen when just playing / pausing / starting the playback or when using structure to navigate. Observed in Chrome and Safari on desktop. Also observed in Safari on iPhone and iPad with an audio item (but only on the second click in the progress bar) as well as on Android Chrome.


https://github.com/user-attachments/assets/5f5dcf51-035e-45c0-b7bb-231ab719d176

Desired behavior would be to remove that highlighting on the progress bard when the user scrubs (clicks in progress bar or drags the playhead).

elynema:
Dananji reverted this design for the audio player embedded widget because the 2 layer approach was causing a vertical scroll in the embeds. So the embedded audio player has the same design as it used to. Both video and audio are now fitting fine in the embedded environment.

Embedded audio seems to be missing the playback speed and audio quality widgets. Will likely need to create a new issue for that.

PR #715, opened 2024-11-04
Use control-bar width to add padding to progress-bar on sides

Dananji:
Related issue: #687

Use control-bar width to add space on either side of the progress-bar in between the player borders and progress-bar. The previous implementation used only CSS by adjusting the width of the progress-bar which was not responding well to player resize.

Screenshot 2024-11-04 at 4 19 11 PM


cjcolvar (approved):
:+1:

Issue #717, opened 2024-11-05
[BUG] Caption icon and captions disappear when switching videos and changing quality


abankhele:
**Describe the bug**
When there are multiple media files, selecting a video from sections other than the first, then returning to a video with captions and changing its quality, causes the caption icon and captions to disappear.

**To Reproduce**

1. Go to 'https://avalon-dev.dlib.indiana.edu/media_objects/3j3332278'
2. Select the second media file (e.g., Venice_5.mp4).
3. Go back to the first media file (e.g., Lunchroom Manners).
4. Use the gear icon to change video quality.
5. Observe that the captions and caption icon are no longer visible.

**Expected behavior**
Changing the video quality should not impact caption visibility; the caption icon should remain visible and captions should remain visible if it is are enabled.

**Screenshots**

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN3VVQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--363af9571ce412ff7bfed7c1c68e9a1112101621/Screen%20Recording%202024-11-04%20at%207.24.09%E2%80%AFPM.mov)

**Desktop (please complete the following information):**
 - OS: macos
 - Browser: Chrome

**Smartphone (please complete the following information):**
 - Device: iPad Pro 12.9 2022 v16.3, iPhone 15 v17.6, Samsung Galaxy Z Fold 6 v14.0
 - OS: ios 17.6, Android, v14.0
 - Browser : chrome, safari

elynema:
Dananji reports that this is a known issue with the quality selector plugin as part of video.js. It has not been resolved on the video.js end, but has been reported.

Second best solution is to trigger event to re-read captions in after quality change.

Dananji:
Done looks like:
Example item to test: https://avalon-dev.dlib.indiana.edu/media_objects/g158bh28p/
- [ ] When captions are turned on, select the second section 
  -> change the quality 
  ->  observe captions icon is in the control bar with green underline 
  -> caption is selected
- [ ] When captions are turned on, select the third section 
  -> change the quality 
  -> observe captions icon is in the control bar with green underline 
  -> the first caption is selected 
  -> select the second caption option
  -> change the quality again
  -> observe second caption is still selected
- [ ] When captions are turned on, select the second section 
  -> change the quality 
[...]

PR #720, opened 2024-11-06
Add back tracks upon quality selection

Dananji:
Related issue: #717 

This is a known issue in the quality-selector plugin we are using. The solution in this PR is to add back in the remote text tracks on quality selection.

### DIFF
--- a/src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
+++ b/src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
@@ -550,7 +550,7 @@ function VideoJSPlayer({
 
       // Enable the first caption when captions are enabled in the session
       if (firstSubCap && startCaptioned) {
-        console.log(firstSubCap.label, firstSubCap.mode);
+        ;
         firstSubCap.mode = 'showing';
         activeTextTrackRef.current = firstSubCap;
         handleCaptionChange(true);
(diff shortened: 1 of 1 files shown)


src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
```diff
@@ -144,6 +147,24 @@ function VideoJSPlayer({
...
+       * This is a known issue with @silvermine/videojs-quality-selector plugin.
+       * When a new source is selected this event is invoked. So, we are using this event
+       * to check whether the current video player has tracks when tracks are available and
+       * adding them back in.
+       */
+      if (tracksRef.current?.length > 0 && isVideo
+        && player.textTracks()?.length <= tracksRef.current?.length) {
+        if (tracksRef.current?.length > 0 && isVideo) {
+          tracksRef.current.forEach(function (track) {
+            player.addRemoteTextTrack(track, false);
```
cjcolvar, line 163: Does this need to de-dupe the tracks when adding?

Dananji: I tested this again in Safari, and it seems removing and adding them back in is a robust solution as Safari handles tracks different than Firefox and Chrome. As the check `player.textTracks()?.length <= tracksRef.current?.length` fails in Safari causing the tracks to duplicate.
Thanks for catching that 🎉

Issue #672, opened 2024-10-22
[BUG] Audio only items do not auto-play next section on Safari (desktop, iPhone)


elynema:
# Description

For audio recordings that have multiple sections, when the first section reaches the end, the second section does not auto-play. Instead, playback stops at the end of the section. This was observed on Safari on desktop and iPhone. In addition, for a playlist where there are multiple audio recordings in a row, autoplay does not advance past the first audio file. 

Video recordings auto-play multiple sections fine. Audio recordings in Chrome are working for auto-play.

Example records for testing:

-  https://avalon-dev.dlib.indiana.edu/media_objects/fn106x94r
-  https://avalon-dev.dlib.indiana.edu/media_objects/gx41mh86p
-  https://avalon-dev.dlib.indiana.edu/playlists/5 (last 4 playlist items are from audio recordings)
  
Event handling _should_ be the same across audio and video and these changes should just be cosmetic

# Done Looks Like
- [ ] Playback continues to the next section/item in Safari when autoplay is enabled

abankhele:
Tested in Safari on both desktop and iPhone for the mentioned items. Everything is working as expected; playback continues when autoplay is enabled.

PR #722, opened 2024-11-06
Trigger ended event when end = duration

Dananji:
Related issue: #672 

When end time equals to duration Safari doesn't emit the `ended` player while other browsers do. Therefore extending the condition to include the cases where end == duration.
Test example manifest: https://avalon-dev.dlib.indiana.edu/playlists/5/manifest auto-advance from 98th item to next.


cjcolvar (approved):
Looks good to me.

PR #726, opened 2024-11-08
Change order of player.load() and setting isReady to keep markers

Dananji:


cjcolvar (approved):

PR #735, opened 2024-11-12
Invoke handleTimeupdate to update markers on player reload

Dananji:
When switching between canvases the fragment markers do not get reset when player is paused. This creates overflowing fragment markers for the progress-bar. This was introduced when `setIsReady(true)` was moved from `loadedmetadata` event to `loadeddata` event to update playlist markers in Safari browsers (related PR: https://github.com/samvera-labs/ramp/pull/726).

Example: Load second section of https://avalon-dev.dlib.indiana.edu/media_objects/3j3332278

 Screenshot 2024-11-12 at 3 23 58 PM


cjcolvar (approved):
:+1:

Issue #723, opened 2024-11-07
[BUG] Zoom on iPad makes widget bar much wider than media player


elynema:
**Describe the bug**
I believe this problem is limited to Safari on iPad. When the text zoom is increased on the browser, the media widget bar in the player does not line up with the video.

This does not happen using the same zoom feature in Safari on iPhone. Also does not happen using the percent zoom in Chrome or Safari on desktop.

**To Reproduce**
1. Go to https://avalon-dev.dlib.indiana.edu/media_objects/3j3332278 in Safari on iPad.
2. Click AA icon just to the left of the URL in the address bar to change text size.
3. Increase to 125%
4. The widget bar in the media player expands horizontally and goes over top of the tabs. The width of the media item itself stays the same.

**Expected behavior**
Widget bar and media should remain same width as text is zoomed and not overlap the tab display.

**Screenshots**
<img width="600" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeU9qQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--587be5277810d11806fcb7edb71e7bf65c263b41/screenshot-iPad%20Pro%2013%202024-17.0.png" alt="screenshot-iPad Pro 13 2024-17.0.png" />

**Smartphone (please complete the following information):**
 - Device: iPad Pro 13
 - OS: iOS17
 - Browser: Safari

joncameron:
Looks good on the Ramp demo using the manifest referenced in the steps to reproduce.

PR #736, opened 2024-11-12
Use resize event on window.visualViewport to adjust control-bar width for mobile devices

Dananji:
Related issue: #723 

**Merge after 3.3.0 tag is created**


cjcolvar (approved):
:+1:

Issue #734, opened 2024-11-12
Alignment Issue Causes Visual Artifacts


joncameron:
# Description
A vertical line appears to the left of the progress bar:
![Screenshot 2024-11-12 at 1 14 19 PM](https://github.com/user-attachments/assets/65fd7588-8bc3-408a-b882-4208ef352d54)

![Screenshot 2024-11-12 at 1 14 12 PM](https://github.com/user-attachments/assets/a35a36e4-9388-4aee-93b4-06e4f71c9c38)

Hover on the progress bar also toggles things in such a way to display a small area of black:
![Nov-12-2024 13-15-35](https://github.com/user-attachments/assets/77f902f0-1fbf-47d6-a145-8f7abcb5ca0a)

# Done Looks Like
- [ ] Element sizing changed to remove black line artifacts from display

joncameron:
Looks good on the Ramp demo site 👍

PR #737, opened 2024-11-12
Remove space between control-bar pseudo elements, to mouse display on hover

Dananji:
Related issue: #734


cjcolvar (approved):

Issue #728, opened 2024-11-08
[BUG] Scrubbing in Safari before playback doesn't work intermittently


elynema:
**Describe the bug**
If the user selects a new section and quickly scrubs someplace fairly late in the video before beginning playback, the playhead will jump back to the beginning of the video.

If the user sits for a few seconds before scrubbing after selecting the section, the scrubbing works. Likely due to the video not being fully loaded.

**To Reproduce**
1. Go to https://avalon-dev.dlib.indiana.edu/media_objects/fq977t794
2. select second section
3. Quickly click on progress bar to scrub to somewhere past the middle of the video.
4. Wait. You’ll see the playhead jump back before starting playback.

**Expected behavior**
Scrubbing should work even before video is fully loaded into the player.

**Screenshots**
[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBenVwQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--9536f4a099715a6105e7c903a9a1d5c35b4f750d/Screen%20Recording%202024-11-08%20at%201.11.21%E2%80%AFPM.mov)

**Desktop (please complete the following information):**
 - OS: iOS 
 - Browser: Safari

Dananji:
We are intentionally calling `player.load()` upon page load for Safari because the player doesn't fully load automatically like in other browsers. This helps us to use scrubbing/seeking on the player progress-bar or use structured navigation and playlist markers to scrub/seek before playback starts.
This is causing the player to reset on Safari, if the user scrubs/seeks within the window when the is revealed and `player.load()` is called.

Dananji:
The fix put in place for this issue could create a race condition for player dispatches for current time updates in the state. I did the following in the demo instance to see whether this happens, and I did not run into any inconsistencies.

But we should probably do these extra tests once this is brought into Avalon, since that environment could create different results.

- Load a new section and seek immediately to a non-zero time and observe the section starts from the seeked time when it is played
  - load https://avalon-dev.dlib.indiana.edu/media_objects/g158bh28p
  - load second section immediately
  - seek to a timepoint > 7 minute mark
  - play the section; observe the player starts from seeked timepoint
- Load a new section and seek immediately to a non-zero time and load another section right after: observe the second section has the time set to zero
  - load https://avalon-dev.dlib.indiana.edu/media_objects/g158bh28p
[...]

PR #741, opened 2024-11-15
Fix seeked time reset on player load in Safari

Dananji:
Related issue: #728


cjcolvar (approved):
This seems fine as far as I can tell although I am wondering if this creates duplicate event handling that could create duplicate `playerDispatch` events leading to potential race conditions.

Dananji:
I didn't run into race condition when testing this out. I'll make a note in the ticket for testing purposes.

Issue #710, opened 2024-11-04
AVAnnotate manifests don't show media resources


Dananji:
### Description

AVAnnotate manifest from the interoperability testing worksheet doesn't load the media into the player.

Links to the manifests:
https://tanyaclement.github.io/znh-1939/manifests/manifest.json
https://avannotate.github.io/SuspenseTheLonelyVilla/manifests/manifest.json

Both these manifest display next/previous buttons, but no message. These don't seem to be restricted items, as they are loading and playing in other viewers.

### Done Looks Like
- [x] Load the media resource from the manifest

elynema:
I did hear there may be some issues with the AVAnnotate manifests themselves, so should run these through the IIIF validator before developing around them.

saracarl:
Yeah, the IIIF validator hangs because it can't handle "range" stanzas.  There's an issue open on the validator for that.

joncameron:
👍

PR #742, opened 2024-11-19
Fix error message display and general parsing error for AVAnnotate manifests

Dananji:
Related issue: #710 

The error message was displaying the in player, but it had a grey overlay on top of it without letting the user access the error message modal. This is because the error message was created by VideoJS instead of the override we've implemented in Ramp.
 
Screenshot 2024-11-19 at 3 32 23 PM

The fix in this PR catches `player.error()` in the `player.ready()` event handler and trigger the custom event handler for player errors. When there is only one source for the Canvas and VideoJS is facing issues playing it, the error display modal dialog is shown as non-dismiss-able modal by setting the following option for VideoJS;
```
errorDisplay: {
  uncloseable: sources?.length > 1 ? false : true,
}, ...
```

In addition to that, this PR adds a check in `manifest-context.js` for playlist manifests before trying to read `highlighting` annotations in a Manifest, as some AVAnnotate manifests fails to load without this check. 
Example manifest from AVAnnotate which had problem loading without this check:
- https://lgsump.github.io/maya-deren-example/meshes-of-the-afternoon/manifest.json
- https://tanyaclement.github.io/znh_jacksonville_1939/s1576-t86-244/manifest.json

This PR needs to be merged after #740, as the changes in parsing in the other PR are needed to load the manifests in this issue.

### DIFF
--- a/src/components/MediaPlayer/MediaPlayer.js
+++ b/src/components/MediaPlayer/MediaPlayer.js
@@ -179,6 +179,11 @@ const MediaPlayer = ({
         sources: isMultiSourced
           ? [sources[srcIndex]]
           : sources,
+        errorDisplay: {
+          // If only one source is available and it is not playable remove 
+          // close button for the error modal dialog
+          uncloseable: sources?.length > 1 ? false : true,
+        },
       } : { ...defaultOptions, sources: [] };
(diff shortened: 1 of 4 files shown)


cjcolvar (approved):
:+1:

Issue #727, opened 2024-11-08
[BUG] Caption text very large on iPad full screen


elynema:
**Describe the bug**
When in full-screen mode in Safari on iPad, the caption text is huge.

This is not happening in full-screen on Android in Chrome, just iPad Safari.

**To Reproduce**
1. Go to https://avalon-dev.dlib.indiana.edu/media_objects/8910jt59d
2. Make sure captions are turned on
3. start playback
4. Go to full screen

**Expected behavior**
Caption text size should be smaller

**Screenshots**
<img width="544" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd0dwQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--eef6183b25e5499741338202221de3d5052eee31/Screenshot%202024-11-08%20at%2012.30.25%E2%80%AFPM.png" alt="Screenshot 2024-11-08 at 12.30.25 PM.png" />

**Smartphone (please complete the following information):**
 - Device: iPad
 - OS: iOS 17
 - Browser: Safari

**Additional context**

elynema:
Note that currently in MCO, the caption text is pretty big in full-screen on iPad:

<img width="599" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeWVwQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--acc49e59c4ebbb819755955b25b143a90c3ca609/screenshot-iPad%20Air%206-17.0%20%282%29.png" alt="screenshot-iPad Air 6-17.0 (2).png" />

This size actually matches most of what I'm seeing on avalon-dev. There was one version of iPad that seemed larger where I grabbed the screenshot in this issue, but I haven't been able to figure out which one.

Dananji:
I tested this on multiple devices for Ramp in `avalon-dev` VideoJS website in Browserstack;

| Device and OS | `avalon-dev` example | `videojs.com` example |
| ------------- | ------------- | ------------- |
[...]

PR #744, opened 2024-11-20
Turn on nativeTextTracks only for browsers using native iOS player

Dananji:
Related issue: #727 

Before:
bs_realios_Tablet_iPad Pro 12 9 2021-17 0 - before fix

After:
bs_realios_Tablet_iPad Pro 12 9 2021-17 0 - after fix

Device tested: iPad Pro 12.9 on iOS 17.1 in BrowserStack


cjcolvar (approved):
:+1:

Issue #724, opened 2024-11-07
Increasing text zoom size on mobile devices does not uniformly increase text size in Ramp components


elynema:
# Description

Testing Ramp in avalon-dev. Increasing the text size in the Safari browser seems to increase text size in only certain Ramp components. Which components are affected seems to differ between iPad/iPhone and Android.

This behavior is not seen when zooming the browser in Chrome or Safari on desktop.

On iPhone or iPad (ios17):

1. Load https://avalon-dev.dlib.indiana.edu/media_objects/3j3332278 in Safari
2. Click the AA icon on the far left side of the address bar
3. Increase text zoom to 150% or 200%
4. Scroll down and you can see that the media object title and sections did increase size, but the Details text did not.

<img width="599" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMWFqQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--797a7e586f57823d807a303a572184bc778fd5b2/screenshot-iPad%20Air%206-17.0%20%281%29.png" alt="screenshot-iPad Air 6-17.0 (1).png" />

On Android Device (Google Pixel 8 or Samsung Galaxy S24, OS 14):

1. Load https://avalon-dev.dlib.indiana.edu/media_objects/3j3332278 in Chrome
2. Click the three vertical dots to the right of the address bar
3. Scroll down and click Settings, then Accessibility. 
4. Slide the slider at the top to 200% to scale text up
5. Go back and reload the web page
6. Scroll down and you can see that the media object title and sections did not increase to 200% size, but the Details and Transcript text did.

<img width="360" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMUdqQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--4bebf282845d2fa1c8b879a4655df2d66e398e39/screenshot-Google%20Pixel%208-14.0%20%282%29.png" alt="screenshot-Google Pixel 8-14.0 (2).png" />

Not sure if this is a bug  or if we need to discuss possible solutions. It does seem like all text should increase in size uniformly to meet the WCAG requirements.

Dananji:
[...]

PR #747, opened 2024-12-03
Set font-size CSS in metadata display component and cleanup

Dananji:
Related issue: #724

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -2,6 +2,8 @@
 
 <head>
   <title>Ramp</title>
+  <!-- Allows for text scaling in the demo instance -->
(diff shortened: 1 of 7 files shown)


cjcolvar (approved):
This looks good as far as I can tell.  With these changes in a number of places would it make sense to make a variable (or set of variables like `$fontSizeLarge`, `$fontSizeMedium`, etc.) or does Bootstrap have variables for this already?

Dananji:
Yes, creating variables for specific font-size values sounds good! I'll work on that.
No, Ramp doesn't use any CSS frameworks at the moment. So, we would defined as sass variables in the `_vars.scss`

cjcolvar (approved):
Looks good to me!

PR #749, opened 2024-12-06
Enable error close button for multiple canvas manifests

Dananji:
Fix a bug introduced in https://github.com/samvera-labs/ramp/pull/742.

Enable close button in error display in the player, when there are multiple canvases.


cjcolvar (approved):
:+1:

Issue #431, opened 2024-02-23
Support multiple motivations for a single annotation


elynema:
**Is your feature request related to a problem? Please describe.**
Annotations in manifests produced by AVAnnotate seem to often contain multiple motivations. In the example below, there is a textual annotation that is labeled as both `supplementing` and `commenting`. 

**Describe the solution you'd like**
It is currently unclear the best solution. In this example, it seems that the `commenting` motivation is the most meaningful. Perhaps discuss with AVAnnotate team the purpose of the additional `supplementing` motivation.

In some scenarios, might we display the annotation twice to respond to both motivations?

**Describe alternatives you've considered**

**Additional context**
Here is an example from this manifest: https://lgsump.github.io/maya-deren-example/meshes-of-the-afternoon/manifest.json 

```
"@context": "http://www.w3.org/ns/anno.jsonld",
      "id": "meshes-of-the-afternoon-canvas-1-shadow-annotation-0.json",
      "type": "Annotation",
      "motivation": [
        "supplementing",
        "commenting"
      ],
      "body": {
        "type": "TextualBody",
        "value": "Here we get the first glimpse at something important to understanding the artist’s play with shadow in that the form already is chosen in Black & White film.",
        "format": "text/plain",
        "purpose": "commenting"
      },
      "target": {
        "source": {
          "id": "https://lgsump.github.io/maya-deren-example/meshes-of-the-afternoon/canvas-1/canvas",
          "type": "Canvas",
          "partOf": [
            {
              "id": "https://lgsump.github.io/maya-deren-example/meshes-of-the-afternoon/manifest.json",
              "type": "Manifest"
            }
          ]
        },
        "selector": {
          "type": "PointSelector",
          "t": "16"
        }
      }
```

# Done Looks Like
- [ ] Email to AVAnnotate members about what they would like to see when there are multiple motivations
- [ ] How are the Aviary player and AVAnnotate rendering multiple motivations for an annotation in a manifest documents?

joncameron:
Right now Aviary is able to display these manifests but Ramp is not. We can look at the Aviary player for how implementation goes, and talk to AVAnnotate project members to get more background on how the supplementing motivation is being used. We can then make changes to support the expectations of rendering when there are multiple motivations.

Dananji:
In the new AVAnnotate manifests, there's an added motivation at the Annotation level where it has motivation: ["commenting", "tagging"].
[...]

PR #751, opened 2024-12-11
Annotation parser

Dananji:
Related issue: #431

### DIFF
--- a/src/services/annotations-parser.js
+++ b/src/services/annotations-parser.js
@@ -7,23 +7,13 @@ export function parseAnnotationSets(manifest, canvasIndex) {
 
   // return empty object when canvasIndex is undefined
   if (canvasIndex === undefined || canvasIndex < 0) {
-    return {
-      ...info,
-      error: 'Error fetching content'
-    };
+    return null;
   }
 
   // return an error when the given Manifest doesn't have any Canvas(es)
   const canvases = manifest.items;
-  if (canvases?.length != 0) {
+  if (canvases?.length != 0 && canvases[canvasIndex] != undefined) {
     canvas = canvases[canvasIndex];
-
-    if (canvas === undefined) {
-      console.error(
-        'iiif-parser -> getMediaInfo() -> canvas undefined  -> ', canvasIndex
-      );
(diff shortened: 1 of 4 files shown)


src/services/utility-helpers.js
```diff
@@ -252,29 +252,40 @@ export function fileDownload(fileUrl, fileName, fileExt = '', machineGenerated =
...
+}
+
+/**
+ * Parse comma seperated media-fragment
+ * @function Util#parseTimeStrings
+ * @param {String} fragment media fragment
+ * @param {Number} duration Canvas duration
+ * @returns {Object} {start: Number, end: Number }
+ */
+export function parseTimeStrings(fragment, duration = 0) {
```
cjcolvar, line 268: Nice refactoring to split this out!

src/services/annotations-parser.js
```diff
@@ -0,0 +1,155 @@
...
+export function parseAnnotationSets(manifest, canvasIndex) {
+  let canvas = null;
+  let annotationSets = [];
+
+  // return empty object when canvasIndex is undefined
+  if (canvasIndex === undefined || canvasIndex < 0) {
+    return null;
+  }
+
+  // return an error when the given Manifest doesn't have any Canvas(es)
```
cjcolvar, line 13: I'm not seeing an error getting returned for this case (looks like it is `{canvasIndex, []}`).
Should this comment be removed or the error added?

cjcolvar (approved):
Looks good to me.  I just had one question about a comment.

src/services/annotations-parser.js
Dananji: Thanks for catching this! The comment should be removed. I have a test case handling this use-case in the other [PR](https://github.com/samvera-labs/ramp/pull/753).

Dananji: @cjcolvar In the other [PR](https://github.com/samvera-labs/ramp/pull/753), it returns `null` instead of `{ canvasIndex, annotationSets: [] }`. I can make this change in there if the latter makes more sense.

cjcolvar (approved):

Issue #437, opened 2024-02-23
Support display of external annotations via an AnnotationPage reference


elynema:
**Is your feature request related to a problem? Please describe.**
AVAnnotate manifests may reference an external AnnotationPage rather than embedding annotations directly into the manifest itself. 

**Describe the solution you'd like**
Ramp should automatically pull in the annotations in the external AnnotationPage to determine if they should be displayed. 

**Describe alternatives you've considered**

**Additional context**
IIIF recipe for embedded or reference annotations: https://iiif.io/api/cookbook/recipe/0269-embedded-or-referenced-annotations/

Dananji:
Example Manifest with linked external `AnnotationPage`: https://lgsump.github.io/maya-deren-example/meshes-of-the-afternoon/manifest.json
External annotation set from linked resource: https://lgsump.github.io/maya-deren-example/annotations/meshes-of-the-afternoon-canvas-1-shadow.json
Annotations in this external annotations follows the format expected in #431. 

Scope of this ticket, would be to identify the linked external annotations in the `AnnotationPage` and then pass them over to the parsing mechanism where multiple motivations are handled.

joncameron:
👍

PR #754, opened 2024-12-16
External annotationpage

Dananji:
Related issue: #437

### DIFF
--- a/src/services/annotations-parser.js
+++ b/src/services/annotations-parser.js
@@ -18,9 +18,10 @@ export function parseAnnotationSets(manifest, canvasIndex) {
     const duration = Number(canvas.duration);
 
     annotationSets = parseAnnotationPage(annotations, duration);
+    return { canvasIndex, annotationSets };
+  } else {
+    return null;
   }
-
-  return { canvasIndex, annotationSets };
 };
 
 /**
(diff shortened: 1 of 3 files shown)


src/services/annotations-parser.js
```diff
@@ -17,23 +17,85 @@ export function parseAnnotationSets(manifest, canvasIndex) {
...
-  return { canvasIndex, annotationSets };
-};
+/**
+ * Parse a annotations in a given list of AnnotationPage objects.
+ * @param {Array} annotationPages AnnotationPage from either Canvas or linked .json
+ * @param {Number} duration Canvas duration
+ * @returns {Array<Object>} a parsed list of annotations in the AnnotationPage
+ * [{ label: String, items: Array<Object> }]
+ */
+function parseAnnotationPage(annotationPages, duration) {
```
cjcolvar, line 82: Since this method is handling multiple annotation pages, how about renaming it?
```suggestion
function parseAnnotationPages(annotationPages, duration) {
```
Should the parsing of a single annotation page be split out so `parseExternalAnnotationPage` can all it?

src/services/annotations-parser.js
```diff
@@ -17,23 +17,85 @@ export function parseAnnotationSets(manifest, canvasIndex) {
...
+};
+
+/**
+ * Fetch and parse linked AnnotationPage json file
+ * @param {String} url URL of the linked AnnotationPage .json
+ * @param {Number} duration Canvas duration
+ * @returns {Object} JSON object for the annotations
+ * 
+ */
+export async function parseExternalAnnotationPage(url, duration) {
```
cjcolvar, line 34: Where does this method get used?  Should it get called in `parseAnnotationPages`?

cjcolvar (commented):
This is looking like a good foundation.  I'm looking forward to the next steps where these annotations will surface in the UI.

src/services/annotations-parser.js
Dananji: This method is supposed to get called from the UI component when a selection is made. I tested it locally with some dummy code in the interface. I didn't commit that code, as it was not complete code plus I didn't want to mix the parsing stuff with the UI.

cjcolvar (approved):

Issue #649, opened 2024-09-23
Add display of annotations to Markers component (UI)


elynema:
# Description

For AVAnnotation grant effort, Ramp needs to be able to display annotations more generically. The Markers component can be re-used for this purpose. The name of the component can be set through a prop.

<img src="https://github.com/user-attachments/assets/5ff32029-14dc-4db2-b25d-096c2c6c3aba" width="50%" height="50%" />
<img src="https://github.com/user-attachments/assets/7516678a-e712-470c-81ea-6d3f7df97767" width="50%" height="50%" />

# Questions
 - [x] How to handle multiple annotation bodies? Do they all get concatenated into a single annotation? Do we want all textualbodies regardless of purpose or just  some?
   - Yes, concatenate with a new line between. Yes to all textualbodies unless purpose=tagging. 
 - [x] When providing a label to filter annotation sets and to display next to each annotation, do we use the AnnotationPage label (Aviary and AvAnnotate provide) or the textualBody with purpose=tagging label?
     - Use AnnotationPage label
 - [x] Annotations with motivation of commenting (AVAnnotate) and supplementing (Aviary) were called out. Are there other motivation types that we need to support?
 - Commenting and supplementing are the primary use cases, and the ones we have been asked to support.

# Done Looks Like

- [x] Extend Markers display component
- [x] Provide a multi-select filter drop-down at the top allowing users to select one or more annotation sets to display. When not expanded, the drop-down should say how many sets are selected.
- [x] Display annotations with motivation commenting (AVAnnotate) or supplementing (Aviary)
- [x] For a given annotation body, the label/value pair displays, ex: "Title: Bladderball"
- [ ] If there are multiple annotation bodies, all should be displayed. They should be concatenated with a new line between
- [x] If there are multiple annotations with the exact same time point / time range, they should concatenated with a new line between
- [x] If there is an annotation textualbody with purpose=tagging, that value should be displayed as the label for the annotation
- [x] Props exist for: component heading, which motivation annotations should be displayed
- [x] auto-scroll checkbox works just like in transcripts
- [x] Time should be displayed as either a single time (if the annotation reference a timepoint or a range if referenced in annotation). If a range, the entire range should link to the starting timepoint in the range.
- [x] Do not display markers on the time rail for first iteration
[...]

PR #758, opened 2025-01-03
Display Aviary annotations and filter linked annotations

Dananji:
- Add support for parsing and displaying Aviary style annotations: 
  - display label/value pairs from the annotation body
  - combine and display multiple annotations for a single time-point as one annotation
- Filter the linked annotation resources in the Canvas to only display time-synced annotations

Related issue: #649


cjcolvar (approved):

PR #760, opened 2025-01-10
Check player is not null before player.trigger('resize') event

Dananji:
When switching manifests in demo site using `Set Manifest` form, the player instance is set to null momentarily. During this window `player.trigger('resize')` is getting called because, the resize event on window is fired when the component tree is cleared when manifest URL in the `Set Manifest` form is emptied or changed.

Adding this check prevents the code from trying to emit the 'resize' event on a null instance of the player causing the page to crash. However this doesn't break the resize functionality, as this function is called again when player instance is properly set.


cjcolvar (approved):
Good catch!

Issue #764, opened 2025-01-24
Fix links in annotation texts


Dananji:
### Description

Annotation texts in a Manifest can sometimes contain links. In the current implementation of annotations component, these links show up as links in the text, but they don't do anything when clicked.

An example Manifest from Aviary: https://yalemssa.aviaryplatform.com/iiif/028pc2tq0m/manifest with such annotation from `Annotation [Transcript]` set is displayed as follows;

<img width="461" alt="Image" src="https://github.com/user-attachments/assets/0c3368ae-50f7-4918-a512-7fe70c79444c" />

In [Aviary interface](https://iiif.aviaryplatform.com/player?manifest=https://yalemssa.aviaryplatform.com/iiif/028pc2tq0m/manifest), clicking on the link on this annotation takes the user to the linked page in the same tab.

### Done Looks Like
- [ ] Seek the player to the associated start time when clicked on the annotation (normal behavior)
- [ ] Take the user to the linked page when clicked on the link in; a new tab / the same tab / a pop up window?
- [ ] Change the colors of the link when active/focused from default values

elynema:
Another example manifest that has wikipedia links: https://tanyaclement.github.io/SGW-SI/john-wieners-at-sir-george-williams-university-the-poetry-series-8-october-1966/manifest.json

elynema:
@Dananji In the example shown in this issue description, when all annotation sets are selected, there are 3 annotations that all begin at 00:00:55.329. When I click on the first annotation (the one that has links in the text), the video player scrubs to the correct starting timepoint, but the Annotations tab atuo-scrolls down until only the third annotation that starts at that timepoint displays. This is jolting for the user. Is it possible for the annotation the user clicked on to remain the highlighteded annotation in the Annotations component?

elynema:
@Dananji Also, what colors should I be seeing for active/focused on the links? It looks like a dark green by default with a lighter green on hover? Did you test color contrast?

Dananji:
I just checked the color contrast, and there's level AAA error for color contrast for the link color and its background for an active annotation;

<img width="724" alt="Image" src="https://github.com/user-attachments/assets/de6763fe-e521-4795-b399-cbad618fed4e" />

I will take a look at it tomorrow.

elynema:
@Dananji Is the issue with jumping to a new annotation related at all to the work to fix links, or is it something unrelated? If it's unrelated, it makes sense to create a new issue. It's possible I only noticed it when testing this issue.

[...]

PR #771, opened 2025-02-10
Handle click on a link in annotation: open URL in same tab without seeking player

Dananji:
Related issue: #764


src/components/MarkersDisplay/Annotations/AnnotationRow.js
```diff
@@ -67,6 +67,12 @@ const AnnotationRow = ({
   const handleOnClick = useCallback((e) => {
     e.preventDefault();
     checkCanvas();
+
+    // Handle click on a link in the text in the same tab without seeking the player
+    if (e.target.tagName == 'A') {
```
cjcolvar, line 72: Does this need an additional check that it isn't an anchor tag link or some other link that refers to the current page?

cjcolvar (approved):
Looks good!

PR #773, opened 2025-02-12
Use showNotes prop in transcript search to filter

Dananji:
Fix bug around transcript search when there are notes in the transcripts and not displayed in the UI with `showNotes=false` prop. 
When feeding transcript cues to the search module use the value of `showNotes` prop to filter the transcript cues accordingly.


cjcolvar (approved):
Looks good!

Issue #768, opened 2025-02-05
Displaying a multiple Annotation tags in Annotations component


Dananji:
### Description

An Annotation can have multiple textual bodies with `tagging` motivation, which are displayed in the Annotations component next to time-range.
For example;
This [AVAnnotate Manifest](https://tanyaclement.github.io/znh-1939/manifests/manifest.json) has annotations with multiple tags which are displaying as expected,

<img width="585" alt="Image" src="https://github.com/user-attachments/assets/cf911587-7a38-4ed0-a603-1731c64cde7c" />

However [this other example Manifest](https://benwbrum.github.io/ohms-to-avannotate-experiment/manifests/manifest.json) also from AVAnnotate created for an OHMS project, which has a higher number of tags per annotation is displayed as follows,

<img width="585" alt="Image" src="https://github.com/user-attachments/assets/49b2ad3d-45e1-400b-b2f0-9346b612ffad" />

In the last AVAnnotate meeting, Aviary folks mentioned this issue with displaying a large number of tags with OHMS projects.

### Done Looks Like
- [x] Display all tags without horizontal scrolling of the component, i.e. a mechanism to display all tags in an accessible and readable format
- [x] Heights of the tags are identical without extra space above or below the texts
- [x] Time ranges are displayed in a single line without wrapping to the next line, i.e. tags don't occupy the space of the time ranges

Dananji:
<img width="633" alt="Image" src="https://github.com/user-attachments/assets/03db70b6-a10d-4fa1-9ff5-29ba0c526352" />
<img width="633" alt="Image" src="https://github.com/user-attachments/assets/42aff8bb-c75e-4f98-86fa-d628d18c54d3" />

elynema:
@Dananji I think the Done Looks Like that you have created for this issue is good. I agree and was thinking that some type of show more / show less will be needed when we have long lists of tags. The basics of how this is working here look fine to me. Elsewhere in the Ramp interface (in the structure), we are using the caret character for expand / collapse (^). I'd be interested in seeing that here instead of the +/- unless you've already tried that and didn't like it.

Dananji:
Yes, It'd be good to have consistency across the components. Thanks for mentioning that!
No, I didn't try a similar icon on this implementation. I will try that out.

Dananji:
Current implementation uses a similar calculation to the `Show more/less` feature in the annotations component. 

[...]

PR #775, opened 2025-02-17
Fix display of longer list of tags with show/hide more button

Dananji:
Related issue: #768


cjcolvar (approved):
Looks good.  I just had a small question that's not a blocker so feel free to merge.

PR #776, opened 2025-02-19
Check for annotations before rendering AnnotationsDisplay

Dananji:
Related issue: https://github.com/avalonmediasystem/avalon/issues/6184


cjcolvar (approved):
Good catch!

PR #793, opened 2025-03-25
Use a lighter color for active cues and annotation rows to fix contrast

Dananji:
Related issue: #764 and https://github.com/samvera-labs/ramp/issues/764#issuecomment-2749402131


cjcolvar (approved):
:+1:

PR #794, opened 2025-03-25
Scroll to clicked annotation when there's multiple annotations in range

Dananji:
Related issue: #764 and https://github.com/samvera-labs/ramp/issues/764#issuecomment-2749165833


cjcolvar (approved):
Looks good to me.

Issue #790, opened 2025-03-24
Show More renders improperly for a specific annotation in Annotations tab


elynema:
**Describe the bug**
The Show More button shows up for certain annotations on the Annotations tab when they are only wrapping to 4 lines of text, not 6.

**To Reproduce**
1. Go to Ramp demo site (https://ramp.avalonmediasystem.org_
2. Set manifest to https://tanyaclement.github.io/SGW-SI/john-wieners-at-sir-george-williams-university-the-poetry-series-8-october-1966/manifest.json.
3. Scroll down to annotation at time point 00:03:15.000.
4. Make browser width less until Show More button appears. Note that Show More appears when there are only 4 lines of annotation displaying, not 6 as usual.
5. For other annotations in this manifest, Show More seems to be working properly.
This also seems to be happening in the annotation at 00:13:38.000.

**Expected behavior**
I would not expect the Show More button to show up in this annotation until the screen width is narrow enough that there are at least 6 lines of text in the annotation.

**Screenshots**

![Image](https://github.com/user-attachments/assets/5b810e6a-465f-4501-aab6-8cc1ce1c47d0)

**Desktop (please complete the following information):**
 - OS: macOS laptop
 - Browser: Chrome and Safari

**Additional context**
Could it be possible that HTML tags (links) in this annotation are causing the browser to think the annotation is taking up more space than it is? If that is the case for this edge case, we may decide not to pursue a complex solution.

joncameron:
The original text is looked at for show more/less functionality, rather than based on the amount of text able to displayed within the current size of the annotation.

elynema:
Note that this is intended to be investigation first. If effort is significant, we will likely de-prioritize this one.

joncameron:
Looks great! Examples of how this looks now:

![Image](https://github.com/user-attachments/assets/c066c9f6-e668-4f40-a178-8e2ec0fec3ad)

![Image](https://github.com/user-attachments/assets/79730dcf-4f91-4bd7-af5f-ba752f080326)

![Image](https://github.com/user-attachments/assets/5ce3371c-5ff7-45af-9b32-c028c4fddab4)

PR #800, opened 2025-04-07
Truncate only text nodes without counting HTML markup, cleanup useEffects

Dananji:
Related issue: #790 

In addition to the fix the PR includes cleanup of `useEffect` hooks in `AnnotationRow` component. There's a bunch of `useEffect` code in the component, which were moved into a custom hook as a result.

### DIFF
--- a/src/services/ramp-hooks.js
+++ b/src/services/ramp-hooks.js
@@ -1453,8 +1453,8 @@ export const useShowMoreOrLess = ({
           }
 
           // Truncate text if the annotation text is longer than max character count
-          const { truncated, hasShowMore } = truncateText(elementText, maxCharactersToShow);
-          if (hasShowMore) {
+          const { truncated, isTruncated } = truncateText(elementText, maxCharactersToShow);
+          if (isTruncated) {
             setTextToShow(truncated);
             setTruncatedText(truncated);
             setIsShowMoreRef(true);
--- a/src/services/ramp-hooks.js
+++ b/src/services/ramp-hooks.js
@@ -15,7 +15,9 @@ import {
 } from './transcript-parser';
 import {
   CANVAS_MESSAGE_TIMEOUT, checkSrcRange, HOTKEY_ACTION_OUTPUT, playerHotKeys,
-  screenReaderFriendlyTime, identifyMachineGen
+  screenReaderFriendlyTime, identifyMachineGen,
+  truncateText,
+  autoScroll
 } from '@Services/utility-helpers';
 import { getMediaInfo } from '@Services/iiif-parser';
 import videojs from 'video.js';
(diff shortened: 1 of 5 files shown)


src/services/ramp-hooks.js
```diff
@@ -1343,3 +1345,249 @@ export const useAnnotations = () => {
...
+                // Do nothing if text length is longer than charsPerLine
+                paddedText.push(line);
+              }
+            }
+            elementText = paddedText.join('<br>');
+          }
+
+          // Truncate text if the annotation text is longer than max character count
+          const { truncated, hasShowMore } = truncateText(elementText, maxCharactersToShow);
+          if (hasShowMore) {
```
cjcolvar, line 1457: Is this check needed given it is happening on line 1471 outside of this function?
If it is needed what do you think about moving this check to the beginning of the function?  I think that would skip unnecessary processing if `hasShowMore` is `false`.

src/services/utility-helpers.js
```diff
@@ -768,3 +768,112 @@ export const groupBy = (arry, key) => {
...
+    return { truncated: htmlString, hasShowMore: false };
+  }
+
+  // Create a temporary div to work with the HTML
+  const tempDiv = document.createElement('div');
+  tempDiv.innerHTML = htmlString;
+
+  const textLength = getTextLength(tempDiv);
+
+  if (textLength <= maxLength) {
```
cjcolvar, line 791: This is an edge case but what do you think about this change to account for the space taken by the ellipsis?  If a text is only 3 characters longer than the max then a user wouldn't have to click the show more button to see the the rest when it could have been rendered initially.

```suggestion
  if (textLength <= maxLength + 3) {
```

cjcolvar (commented):
This looks like a good improvement!

Probably outside the scope of this PR but have you thought about truncating on word boundaries instead of mid-word?

src/services/ramp-hooks.js
Dananji: The check on line 1471 is to check whether show more/less is enabled in the props, while this check is to see whether the text is truncated or not. And based on this check show more button is displayed next to the truncated text.
I'm not clear on how this could be moved up, can you please clarify?

Dananji:
> Probably outside the scope of this PR but have you thought about truncating on word boundaries instead of mid-word?

Yes, that sounds great! If it's not complex maybe we can add it to this PR? I can check on that.

src/services/ramp-hooks.js
cjcolvar: Nevermind!  I got confused with the variable names `hasShowMore`  vs.`enableShowMore`.

Dananji: Maybe I can change the variable names to make it clearer? 🙂

cjcolvar: I think it is makes sense as they are.  I just missed that they were different.

src/services/ramp-hooks.js
```diff
@@ -1343,3 +1345,249 @@ export const useAnnotations = () => {
...
+              } else {
+                // Do nothing if text length is longer than charsPerLine
+                paddedText.push(line);
+              }
+            }
+            elementText = paddedText.join('<br>');
+          }
+
+          // Truncate text if the annotation text is longer than max character count
+          const { truncated, isTruncated } = truncateText(elementText, maxCharactersToShow);
```
cjcolvar, line 1456: :+1:

src/services/utility-helpers.js
```diff
@@ -768,3 +768,122 @@ export const groupBy = (arry, key) => {
...
+  // Truncate text nodes
+  if (node.nodeType === Node.TEXT_NODE) {
+    if (node.textContent.trim() === '') return 0;
+    if (node.textContent.length <= maxLength) {
+      return node.textContent.length;
+    } else {
+      // Get the index of the last space character in the truncated text with maxLength
+      const lastSpaceIndex = node.textContent.substring(0, maxLength).lastIndexOf(' ');
+      // If text doesn't have spaces, truncated at maxLength (this is probably an edge case?)
+      // FIXME:: Maybe there's a better way to handle this than breaking the word?
```
cjcolvar, line 842: If the node would be truncated and there is no space, should this truncate the entire textual content of the node?  That might work for English but probably not for Chinese or Japanese and other languages without required spaces.

For example, the last test in utility-helpers.test.js has to truncate mid-word because it only has one word in the node.

What's currently in this PR is probably good enough for now since it improves upon what was there before.  :slightly_smiling_face:

cjcolvar (approved):
Looks good!

Issue #788, opened 2025-03-24
Click anywhere outside drop-down to close multi-select drop down on Annotations component


elynema:
# Description

Currently, after the user opens the multi-select drop-down on the Annotations component to see the list of annotation sets, the drop-down can only be closed by clicking on the triangle collapse button. This seems limiting.

I think this would work better for users if clicking anywhere outside the multi-select drop-down automatically closed it.

https://github.com/user-attachments/assets/dac2ddfa-87c2-4276-9a5a-20afbe2eab56

# Done Looks Like
- [ ] Clicking/selecting outside of the drop-down when active closes the drop-down menu

joncameron:
This isn't using drop-down element native to HTML, so other changes might be required to make this work.

joncameron:
👍

PR #801, opened 2025-04-07
Add event listener for click outside event to close annotations dropdown

Dananji:
Related issue: #788


cjcolvar (approved):
Looks good to me!

Issue #798, opened 2025-04-02
Ramp errors with white screen for specific manifest


elynema:
**Describe the bug**
When entering a specific manifest into Ramp, I get a white screen (as opposed to some type of error message about the manifest). 

**To Reproduce**
1. Go to Ramp demo site (https://ramp.avalonmediasystem.org)
2. Set manifest to https://weareavp.aviaryplatform.com/iiif/hm52f7jz70/manifest.
3. See a white screen (Ramp demo interface disappears)

From the dev panel, it looks like the manifest is being requested a second time after returning successfully the first time, but  them the second request is canceled and Ramp is left in limbo?

**Expected behavior**
Manifest data should be loaded.

**Screenshots**
<img width="1375" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBejZGQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--4cec99c74232e624dce3c7bbdce5fe2f54f8fa9f/image.png" alt="image.png" />

**Desktop (please complete the following information):**
 - OS: macOS laptop
 - Browser: both Chrome and Safari

joncameron:
No items list in the manifest for this example, which is causing the error.

Dananji:
The Manifest in this ticket seem to be fixed by Aviary at the time of development. So, I created an example Manifest (a Manifest without `items` property) to test this in Ramp and fixed the issue we were seeing with the white screen.

PR #802, opened 2025-04-08
Handle manifest w/o items error in ErrorBoundary

Dananji:
Related issue: #798


cjcolvar (approved):
:+1: Looks great!  I like how this should hopefully keep ramp from dying when it encounters errors since the ErrorBoundary is wrapping everything.

Issue #792, opened 2025-03-24
Show less does not take user back to top of long annotation on Annotations tab


elynema:
**Describe the bug**

In this particular example, the annotation is quite long. When it is closed using the show less button, the user is thrown down into a further timepoint in the set of annotations, rather than see the annotation that had been opened.

**To Reproduce**
1. Go to Ramp demo site (https://ramp.avalonmediasystem.org)
2. Set manifest to https://yalemssa.aviaryplatform.com/iiif/028pc2tq0m/manifest
3. Select only the annotation set labeled 'Google Translate Spanish translation of transcript [Transcript]"
4. The first annotation has time range 00:00:00.000 - 00:00:00.000. This annotation concatenates numerous annotations scattered throughout the manifest with this timerange.
5. Click Show More. The annotation is quite long and will require scrolling to see it all.
6. Once you've scrolled to the end of the annotation, click Show less.
7. The annotation collapses. You are left viewing annotation with time range 00:01:00.433 - 00:01:01.433, and you have to scroll back up to get to the original annotation.
It seems likely this is the standard behavior of Show more / Show less, but it's usually not noticeable because the overall annotation is shorter.

**Expected behavior**
User should be taken back to the original annotation in the Annotations tab after clicking Show less.

**Screenshots**

https://github.com/user-attachments/assets/1179380b-269e-4ea1-be9f-89a4a92ed2d5

**Desktop (please complete the following information):**
 - OS: MacOS laptop
 - Browser: Chrome, Safari

**Additional context**

joncameron:
"Show less" now brings the current annotation block to the top of the component as expected, instead of shifting it out of view.

PR #803, opened 2025-04-08
Scroll annotation into view when show less is clicked

Dananji:
Related issue: #792


cjcolvar (approved):

PR #804, opened 2025-04-11
Add back missing autoScroll import

cjcolvar:
The `autoScroll` import was removed in #800 because it wasn't used but #803 was cut from main before #800 was merged and made use of `autoScroll`.  #800 was merged and #803 didn't have any git merge conflicts but the needed import was lost in the process.


Dananji (approved):
🎉

Issue #789, opened 2025-03-24
Auto-scroll and highlighting ignore single timepoints in Annotations tab


elynema:
**Describe the bug**
The new Annotations tab will display annotations that are time ranges, but also that are single time points. If the annotation is a single time point, it does not get highlighted as the time progresses. Since it doesn't get highlighted, the user also does not see any progress through and the content in the Annotations tab does not auto-scroll.

**To Reproduce**
1. Go to https://ramp.avalonmediasystem.org
2. Load manifest: https://tanyaclement.github.io/znh_jacksonville_1939/s1576-t86-244/manifest.json
3. Scroll down to about 47 minutes in and start playback.
2. Notice there is no highlighting or auto-scrolling from 00:47:23.957 to 00:49:43.887 when we get the first range annotation again. There are several time point annotations in this timeframe

Another example with lots of timepoint annotations is https://tanyaclement.github.io/SGW-SI/john-wieners-at-sir-george-williams-university-the-poetry-series-8-october-1966/manifest.json

**Expected behavior**
Not sure of the best solution. Could we set a very brief highlighting for single timepoint annotations like 0.1 or 0.5 seconds? And auto-scroll should continue to move, even though the annotations are not highlighted for a long period of time?

## Done Looks Like
- [ ] Auto-scroll works to make sure the current time stamp or range is at the top of the annotation display
- [ ] Investigation of how feasible it would be to highlight these portions as well in addition to moving them for autoscroll, as if they were time ranges

joncameron:
One solution is to have calculate the gap between timestamps without duration. This would resolve the autoscrolling issue.

Aviary is only autoscrolling on ranges; keeping the current/last available timepoint at the top of the display would be one way to keep this. That would be the first thing to try to implement to make sure that functionality is there.

In this example, they are all individual timepoints: https://tanyaclement.github.io/SGW-SI/john-wieners-at-sir-george-williams-university-the-poetry-series-8-october-1966/manifest.json. This looks like we probably should be highlighting portions that have timestamps but not time ranges, with the expectation that the range is implicit in the space between timestamps, and the annotation content is referring to that implicit range rather than at that exact timepoint.

joncameron:
The component now highlights and auto-scrolls as expected for the manifests in the description.

PR #806, opened 2025-04-16
Auto scroll time-point annotations by comparing start time avoiding end times

Dananji:
Related issue: #789 

Assumption: for time point annotations consider an implicit time range spanning from its given start time to the start time of the next annotation in the list. Therefore marking a time point annotation as active until the current time of the player reaches the start time of the next annotation's start time.


cjcolvar (approved):

PR #816, opened 2025-05-15
Fix double focus in strucutred nav on initial Shift+Tab

Dananji:
Related issue: #809


cjcolvar (approved):
Looks fine as far as I can tell.

Issue #817, opened 2025-05-16
a11y - User can get trapped with TAB in annotation in Annotations component


elynema:
# Description

A user can use keyboard navigation, with or without a screen reader, to navigate a set of annotations in the Annotation component in Ramp. If the user attempts to use TAB to navigate to links within the Annotation component, they can become trapped. 

Example manifest with links in annotations: https://yalemssa.aviaryplatform.com/iiif/028pc2tq0m/manifest

For example, the user may 

1. use the TAB key to move through the page to the set of tab headers
2. use -> key to navigate to the Annotations header
3. use the TAB key to enter the Annotations tab panel and move through the annotation set drop-down and auto-scroll buttons
4. Use the up/down arrows to navigate through annotations
5. Use the TAB key to tab to an annotation with links. 
6. Note that once the user enters an annotation with links, continually pressing the TAB key will cycle through the links within that annotation. The user will not be able to escape back up to the tab panel with SHIFT+TAB or to the next tab element on the page with TAB
7. Instead, the user can use the up/down arrow keys to move to the next annotation. Once they have no longer selected an annotation with links, they can use SHIFT+TAB to escape back to the tab panel.
8. They will never be able to move to the next tab element on the page using TAB until they use SHIFT+TAB to back up to the tab header element.

See discussion on https://github.com/samvera-labs/ramp/pull/805.

elynema:
@joncameron I started this issue, but the behavior I'm currently seeing on the Ramp demo site doesn't seem to match what I saw when I was testing before, so I may need to wait until pull request 805 is merged and then re-test to get it right. https://github.com/samvera-labs/ramp/pull/805

joncameron:
No longer reproducible 👍

Issue #815, opened 2025-05-09
[A11y] VoiceOver doesn't annouce instructions for a selected annotation


Dananji:
### Description

When using keyboard navigation along with VoiceOver in the transcript component, it announces the cue content and instructions to activate a selected cue. The annotations component follows a similar pattern and is expected yield similar behavior, but that seems to be not the case.

Comment highlighting this issue: https://github.com/samvera-labs/ramp/pull/805#issuecomment-2867578837 by Emily;

> When I tab into the Annotation list from the menu and use the up/down arrow keys to select an annotation, voiceover reads it at least twice in a row. That's problematic.
    In the transcript tab, when I use the up/down arrow keys to select a cue, voiceover reads it once and then provides an instruction to use Ctrl + Option + Enter to select it. Doing that seeks to that cue in the player. It looks like the same keyboard combination works to seek to the beginning timepoint for an Annotation in the annotations tab, but voiceover doesn't announce the instructions.

### Done Looks Like
- [ ] VoiceOver reads out content and instructions relevant to a selected annotation row similar to a transcript cue
- [ ] Keyboard navigation works as expected for annotations component (https://github.com/samvera-labs/ramp/issues/781#issuecomment-2806990134)

Dananji:
Changes made to fix this issue:
- use `role="button"` on the focus-able portion of the annotation row, which enables the instruction announcement for activation of the focused annotation by a screen reader

This is waiting on #805 to be reviewed and merged.

Dananji:
This is an easy fix. I can make this change and publish a patch release of Ramp.

Issue #811, opened 2025-05-08
[A11y] Fix role relationship error in Annotations component


Dananji:
### Description

Error description from SiteImprove plugin;

> Role with implied hidden content has keyboard focus: Some elements have a role that implies their internal structure should be hidden from assistive technologies. In other words, their sub-elements have a presentational role.
If a sub-element has keyboard focus, it should be moved outside of the presentation role.

This is referring to the annotations with links in the text. In the current implementation, annotations component is displaying annotations as list elements with `role='option'`. And any children (links within annotation texts) of such an element are considered presentational.
Relevant success criterion: https://www.w3.org/TR/WCAG/#name-role-value (Level A)

### Done Looks Like
- [ ] Identify a suitable role for the annotations elements and replace the existing role with it
- [ ] Access the accessibility design pattern adapted for keyboard navigation is viable

### Note
This error is introduced in the work done in #781, which has an open PR for implementing keyboard navigation for the annotation component. If/when this work is merged to `main`, we will see this accessibility issue in the annotations component.

Dananji:
Changes made to fix this issue:
- replace `<ul>/<li>` elements with `<div>` as the list related HTML elements implicitly add list-based role relationships to the DOM structure, which doesn't allow annotation rows to present links/buttons
- only the timestamp portion of the annotation is presented as a button. With this, the keyboard navigation will only focus on the timestamp portion when using `ArrowUp` and `ArrowDown` keys to traverse the annotation content
  - additionally, add `aria-label` attribute with time and text to this time-tag `<div>` element in the annotation row to be read by screen readers when focus is on the annotation timestamp. 

This work is waiting on #805 to be reviewed and merged.

joncameron:
This is fixed and no longer reported in SiteImprove.

PR #819, opened 2025-05-19
Fix role relationship error and VoiceOver instruction issue for annotation row

Dananji:
Related issues: #815, #817, and #811

Changes in the PR:
- remove `<li />` elements for annotation-row which implicitly adds list based aria-role blocking the use of `role="button"` on annotation row elements. Having `role="button"` let screen-readers to announce instructions for activating the annotation upon focus (#815)
- replace `<ul />` with `<div />` element to remove implicit role relationship associated with the `<ul />` (#811)
- use aria-role 'button' only on the timestamp portion of the annotation row to enable nesting `<a />` and `<button />` tags within annotation texts (#811)
- add `aria-label` to annotation's timestamp, with more user friendly time formats
- some cleanup on the player hotkeys function for changing roles and test everything works

These need to be broken into multiple PRs once the base PR is merged.

Merge after #805

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "@testing-library/dom": "^10.4.0",
     "@testing-library/jest-dom": "^5.11.4",
     "@testing-library/react": "^16.0.1",
+    "@testing-library/user-event": "^14.6.1",
     "@types/jest": "^26.0.12",
     "@types/react": "18.3.5",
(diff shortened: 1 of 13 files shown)


Dananji:
TODO:: active annotation display doesn't work properly when time-range annotations are displayed

PR #820, opened 2025-05-19
Change role relationship in transcript cues

Dananji:
Change transcript roles to fix role relationship error in SiteImprove when using `role="list"` on parent container while each cue is being presented with a `role="button"`.
In addition, add an `aria-label` for each cue with a more user friendly time format for screen readers to use.


cjcolvar (approved):
Looks good!

PR #822, opened 2025-05-23
Reveal player after initial seek event, set currentTime in play promise

Dananji:
Related issue: https://github.com/avalonmediasystem/avalon/issues/6264

### DIFF
--- a/src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
+++ b/src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
@@ -188,11 +188,6 @@ function VideoJSPlayer({
         });
       }
     });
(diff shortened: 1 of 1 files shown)


cjcolvar (commented):
I tested this on a local demo instance (`yarn demo`) using https://avalon-staging.dlib.indiana.edu/media_objects/sj139195q/manifest.json and was still able to reproduce the issue very infrequently in Firefox with the player playing.  I wasn't able to reproduce it with the player paused.

Dananji:
@cjcolvar I was not able to reproduce the issue you were seeing on Firefox in my Mac with the same item. I tested it in both a normal window and a private window. I wonder what's going on?

PR #823, opened 2025-05-27
Fix voiceover and role relationship errors for annotations

Dananji:
Related issues: #811 and #815

To fix the role relationship error in annotations component, `<ul />` and `<li />` elements replaced with `<div />` elements when creating the annotations list. Because these list elements are document structure elements, implicit roles were being added to the DOM. And these roles were blocking the use of `role="button"` for annotations and displaying links and buttons within the annotation text.

As a result of this, #815 gets automatically fixed because, removing `<li>` in the annotation row implementation eliminated the list-based role overriding `role="button"` for a clickable annotation row, which automatically enables the instructions to activate the annotation in VoiceOver 

Additionally;
- added `aria-label` to annotation row with more user friendly time formats
- some cleanup on the player hotkeys function
- fix highlights for active annotation with time ranges


cjcolvar (approved):

PR #824, opened 2025-05-28
Limit button role for timestamps, to avoid a11y issues with links in annotations

Dananji:
Related issue: #817 

Limit ARIA role "button" to annotation's timestamp, allowing annotation text to display links and buttons without breaking accessibility. This change allows all links and buttons in the annotations display to be included to page's natural tab order allowing the user to continue with keyboard tab navigation without getting trapped inside the annotations component.

Additionally, this PR separates the click events on links within annotation texts in a separate event handler. This event handler provides links validation, and enables navigation in the same page when a link is valid.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "@testing-library/dom": "^10.4.0",
     "@testing-library/jest-dom": "^5.11.4",
     "@testing-library/react": "^16.0.1",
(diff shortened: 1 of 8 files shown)


cjcolvar (approved):
Looks good to me! :tada:

PR #825, opened 2025-05-28
Fix file download icon placement on Canvas change

Dananji:
When switching between canvases the file download button gets added before the previous button in the player's control bar
This happens when switching from a Canvas without rendering files to a Canvas with rendering files;

https://github.com/user-attachments/assets/81b1a2ce-b2bc-4fb7-be4e-c742a0dea597


cjcolvar (approved):
Good catch!

Issue #814, opened 2025-05-09
[A11y] Use Landmark roles and/or sectioning HTML elements for easy page navigation


Dananji:
### Description

Apple's VoiceOver provides a web rotor for easy navigation of the DOM structure. It is a feature that allows users of VoiceOver screen reader to navigate web pages more efficiently by accessing lists of structural elements like headings, links, forms, and more. One such structural element category is landmarks, which allows the user to easily access a part of the page without having to tab through all the elements that comes before it.

There are 2 possible techniques to implement landmarks in the DOM structure;
1. Using landmark roles (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/landmark_role)
2. Using sectioning HTML elements (https://www.w3.org/WAI/ARIA/apg/practices/landmark-regions/)

Since Ramp is implemented as a component library we can explore 2 possible ways to implement landmarks. Using the above techniques (one of them or both as needed);
1. At each component level. This would hopefully provide landmarks out of the box for any one who is using Ramp in their application (needs to be tested)
2. At the time components are laid out in an application that is using Ramp. We can try this out in the demo site implementation.

### Done Looks Like
- [ ] Investigate and document how the suggested techniques are used. Can they be used independently or do they need to be implemented together to achieve the expected behavior?
- [ ] Investigate and document at which level it makes sense to implement landmarks. Inside the exported components or at the time of component usage in an application?
- [ ] Implement the code as needed based on the findings
- [ ] Test using VoiceOver to confirm landmarks are being populated appropriately in web rotor

Dananji:
To populate landmarks on the page
- use HTML sectioning [elements](https://www.w3.org/WAI/ARIA/apg/practices/landmark-regions/#htmlsectioningelements) along with `aria-label` attribute to explicitly name regions on the page. For an example, if there are 2 navigation systems in the page using `<nav />` element along with `aria-label` with populate 2 landmarks for VoiceOver web rotor with the associated labels.
- use `role` with an ARIA role that is a subclass of the `landmark` role, which is an abstract role. `role="landmark"` SHOULD NOT be used. Subclasses available within `landmark` role are; 
  - `section`
  - `banner`, instead of this use `<header />`
  - `complementary`
  - `contentinfo`, instead of this use `<footer />`
  - `form`
  - `main`, instead of this use `<main />`
  - `navigation`, instead of this use `<nav />`
  - `region`
  - `search`

[...]

PR #827, opened 2025-05-29
Implement sectioning elements/landmark roles for page navigation

Dananji:
Related issue: #814


demo/app.js
```diff
@@ -134,29 +134,29 @@ const Tabs = ({ tabValues, manifestUrl }) => {
   });
 
   return (
-    <section className="tabs-wrapper">
-      <div className="switcher">
+    <section className='tabs-wrapper' role='navigation' aria-label='more Ramp components in tab'>
```
cjcolvar, line 137: Does having the same `aria-label` here as in the child `ul` cause it to be read twice?  If so, then I could imagine it might be confusing since they have the same label.

cjcolvar (approved):
This looks great as far as I can tell!  I didn't test it out with a screen reader so we'll have to do that as part of release testing.

PR #829, opened 2025-05-30
Bug fix: keep original order of timespans for playlist

Dananji:
While testing the code for 4.0.0 release, I noticed that playlists structure scroll to the item that has the lowest start time on page load, even though the player shows the first item on the page. Example: https://avalon-dev.dlib.indiana.edu/playlists/5, the structure scrolls to the playlist item at the 98th position even though the player is still on the first item.
This was a bug introduced in the work that was done to fix the display of overlapping timespans in the player.


Dananji:
Yes, that is true! I hadn't really thought of that scenario. I'll create an issue to re-visit this post-release.

Issue #836, opened 2025-06-20
Metadata Headers in VTTs Cause Transcripts Tab to Be Invalid


markpbaggett:
**Describe the bug**
We've started adding embedded metadata to our VTTs according to [FADGI guidelines](https://www.digitizationguidelines.gov/guidelines/FADGI_WebVTT_embed_guidelines_v0.1_2024-04-18.pdf) to help track provenance of AI generated captions and subtitles.

Here is an example:

```
WEBVTT
Type: caption
Language: eng
Responsible Party: US, Texas A&M University Libraries
Originating File: /Users/mark.baggett/code/whisper_to_vtt/test_files/rfta_art_73.mp4
File Creator: Baggett, Mark P.
File Creation Date: 2025-06-20 10:56:44.008719
Local Usage Element: [version history] 2025-06-20 10:56:44.008735: WebVTT initially generated by OpenAI Whisper v. 20240930 and model 'tiny' with tamu_whisper v. 0.1.1.

00:01.540 --> 00:10.320
One of the stories that I heard the woman knew that the fire was getting closer
because her cat was sitting next to the vent by the door because the cat could

00:10.320 --> 00:18.940
smell the smoke coming in and that inspired her to leave. And I was just like,
wow, listen to your pets, you know, come to the slaughter than we are. And the

```

If I remove the header from the VTT, [the manifest](https://ramp.avalonmediasystem.org/?iiif-content=https://markpbaggett.github.io/static_iiif/manifests/tests/tamuwhisper-video-two.json) works.

If I don't remove it, I get [Invalid WebVTT file, please check again.](https://ramp.avalonmediasystem.org/?iiif-content=https://markpbaggett.github.io/static_iiif/manifests/tests/tamuwhisper-video.json).

**To Reproduce**
1. Pass a manifest to RAMP with embedded metadata in the header. If you need one, [this](https://ramp.avalonmediasystem.org/?iiif-content=https://markpbaggett.github.io/static_iiif/manifests/tests/tamuwhisper-video.json) should work.
2. Click on "Transcripts"
3. See error

**Expected behavior**
The embedded metadata is ignored but the transcripts tab works as intended.

**Screenshots**

<img width="1073" alt="Image" src="https://github.com/user-attachments/assets/e570ad5c-1fbe-4638-aed7-ee65f0e85939" />

<img width="1073" alt="Image" src="https://github.com/user-attachments/assets/1d8d929f-b563-4476-b4e9-1f4153919ea5" />

[...]

PR #844, opened 2025-07-07
Change WebVTT validation to allow files with embedded metadata

Dananji:
Related issue: https://github.com/samvera-labs/ramp/issues/836

When `showMetadata=true` and the WebVTT file has embedded metadata in the header it is displayed as follows;
Screenshot 2025-07-07 at 10-38-34 Ramp

Example Manifest:https://markpbaggett.github.io/static_iiif/manifests/tests/tamuwhisper-video.json


cjcolvar (approved):

Issue #840, opened 2025-06-25
Can Transcript component open links in a new tab explicitly?


elynema:
# Description

For the purposes of the EVIADA content that IU is migrating into Avalon, we will be creating .vtt files to display in the Transcript component that will have links in each cue to the larger collection description. Currently, we expect these links to point to a supplemental file that is an html file. When testing this, the html file always downloads to the desktop, rather than opening in a new tab.

Is it possible for the Transcript component to manage the behavior of the link somehow so that these HTML files open in a new tab?

Example record: https://avalon-dev.dlib.indiana.edu/media_objects/12579s29x

joncameron:
If we can handle this behavior in Avalon (see related issue), it makes sense to do there and only add to Ramp if needed.

PR #849, opened 2025-07-17
Browser determined download for files tab

cjcolvar:
Part of https://github.com/avalonmediasystem/avalon/issues/6399
Related to #840 

This PR removes the special handling of links from the SupplementalFiles component allowing the server serving those urls to supply appropriate content disposition headers that would allow the file to be rendered inline in the browser instead of forcing download.

It appears that the `download` attribute on anchor tags makes Chrome and Firefox download ignoring the inline content disposition header if present.  See [final note in MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a#download).


Dananji (approved):
This looks good 👍 

As for the build files, I like not including build files in PRs because it makes
- the PR cleaner
- the merging process easier when there are multiple PRs, which shouldn't be an issue with this

And I'm okay with merging this as it it.

If it helps, what I do when I need to test things in Avalon I run `yarn build` and then `yarn pack` to create a `.tgz` file, which I then move to `avalon` workspace and install with yarn using `yarn add ./samvera-ramp-v4.0.0.tgz`

cjcolvar:
Thanks, I forgot about the `yarn pack` process.  That's definitely cleaner.

Issue #847, opened 2025-07-16
[BUG] Clicking through results of transcript component search sometimes doesn't work


elynema:
**Describe the bug**
Testing with our EVIADA content, which is bringing in .vtt files with long texts of chunk in each cue seems to have turned up some kind of bug with search in the transcript component. 

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

Dananji:
This is happening only for some searches, not for all. 
Some of the searches that work as expected: `ceremony`, `orichas`, `akpones` (these are some random words I tried from the transcript).

As for the search query `religious`, there are some items in the search response with `target` values without media fragment information. And looking at the WEBVTT file I came across a couple of cue blocks as follows;
```
04:55:33.914 --> 04:58:17.423
<b>Entrance into Templo San Benito through crowd during tambor</b>
Cuban folklorist Ernesto Armiñán Linares leads American researcher Kristina Wirtz and her husband Ed Eckel (videographer) into the crowded temple-house Templo de San Benito de Palermo during a bembé or festive drumming ceremony for Saint Barbara held on the Eve of Saint Barbara's Day. Singer Tarabán of the Reynerio Pérez family, which owns the Templo San Benito, is the lead singer, and conga drums provide the accompaniment.
<b>lead singer</b>: Bordele Pérez a.k.a Tarabán

<b>Social & Cultural Groupings</b>: Singers, Santeros (Priests), Religious gatherings, Drummers (Musicians)
<b>Genres & Performance Types</b>: Sacred music, Ritual, Santeria music, Religious dance, Sacred songs, Santeria, Rites and ceremonies
<b>Languages</b>: Lucumi
<b>Instruments</b>: Conga (Drum), Rattle (Musical instrument), Cowbells
<b>Venue Types</b>: Shrines (structures)
<a href='https://avalon-dev.dlib.indiana.edu/media_objects/12579s29x/supplemental_files/488'>See full annotation details</>
```
where there's an empty line inside the cue block, which is not permitted in the WEBVTT spec. I checked this in [W3C's WEBVTT validator](https://w3c.github.io/webvtt.js/parser.html) and it was identified as an invalid cue.

[...]

PR #852, opened 2025-07-23
Remove untimed search hits and preserve inline styling for search within

Dananji:
Related issue: #847 

Changes:
- apply highlights from search response to the styled transcript text in Ramp to preserve styling specified in the transcript file. This change is specifically for timed text in WEBVTT files
- remove items in the search response without media-fragment information for timed transcripts


cjcolvar (approved):
This looks good as far as I can tell.

PR #863, opened 2025-08-21
Set player time accurately when switching between canvases

Dananji:
Related issue: https://github.com/avalonmediasystem/avalon/issues/6531


cjcolvar (approved):
Looks okay as far as I can tell.

Issue #857, opened 2025-07-30
[BUG] Display inline annotations from Aviary manifests


Dananji:
### Description

Transcript component does not display annotations with `supplementing` motivation for Aviary manifest:  https://weareavp.aviaryplatform.com/iiif/hm52f7jz70/manifest

This Manifest has 5 annotation sets in the first Canvas, where 2 of them have the `supplementing` motivation, which should be displayed in the `Transcripts` component. The `Annotations` component however displays all 5 annotation sets.

### Done Looks Like
- [ ] Inline annotations with `supplementing` motivation are displayed in Transcripts component

Dananji:
With this change, `Transcript` component will be able to display inline annotations with `supplementing` motivation. However, it will not display any linked annotation pages (.json files) if there are no specified motivations. An example would be annotations in https://tanyaclement.github.io/znh_jacksonville_1939/s1576-t86-244/manifest.json.

PR #864, opened 2025-08-25
Display inline supplementing annotations in transcript component

Dananji:
Related issue: #857

### DIFF
--- a/src/services/transcript-parser.js
+++ b/src/services/transcript-parser.js
@@ -51,6 +51,8 @@ export const TRANSCRIPT_CUE_TYPES = {
   metadata: 'METADATA'
 };
 
+export const TRANSCRIPT_MOTIVATION = 'supplementing';
+
 /**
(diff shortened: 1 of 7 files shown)


src/services/transcript-parser.js
```diff
@@ -371,13 +389,33 @@ export async function parseTranscriptData(url, format, canvasIndex, parseMetadat
...
+ * @param {Array} annotations 
+ * @returns {Array}
+ */
 function createTData(annotations) {
   if (annotations?.length === 0) return [];
   let tData = [];
+
+  // Build text from an array of TextualBody in an annotation
+  let buildText = (texts) => {
+    if (!texts?.length === 0) return '';
```
cjcolvar, line 404: I don't understand the logic in this line.  When I test in the browser console I don't see how this can ever be true.  Can you please explain?

cjcolvar (approved):
Looks good overall!  I just had one question which isn't a blocker.

cjcolvar (approved):

PR #865, opened 2025-08-26
Fix close/expand for cookbook ToC recipe 0065

Dananji:
The close/expand functionality was broken for a [IIIF Cookbook A/V ToC recipe](https://ramp.avalonmediasystem.org/?iiif-content=https://iiif.io/api/cookbook/recipe/0065-opera-multiple-canvases/manifest.json), because the format of its `structures` property sets the conditional rendering of the section's children to always be displayed breaking the close/expand functionality.


cjcolvar (approved):

Issue #884, opened 2025-11-07
[BUG] Fix Transcripts/Annotations Synchronization for multi-item Canvas


Dananji:
### Description

The [IA Manifest](https://iiif.archive.org/iiif/CSPAN3_20180217_164800_Poplar_Forest_Archaeology/manifest.json) is a single Canvas Manifest with multiple items in the same Canvas. Similar to [this cookbook recipe](https://iiif.io/api/cookbook/recipe/0064-opera-one-canvas/).
This particular IA Manifest has a `supplementing` annotation that is displayed in the transcripts and annotations components. The cues/annotations are created with time-stamps relative to the entire duration of the Canvas (all 13 items combined). 
However transcripts and annotations components are not aware of this multi-item Canvas. This breaks cue/annotation synchronization with playback, because the transcripts/annotations components don't consider the relative start time of each item in the Canvas, which it should.
Therefore, when a user clicks on a cue/annotation it seeks the player to the timestamp relative to the current media item's start point instead of the entire duration.

Steps to reproduce:
1. Load the Manifest into Ramp
2. Select the `Transcripts` tab and uncheck `Auto-scroll with media`
3. Click on `[00:01:20] LANDSCAPE,DERSTAND` cue
4. Observe that, the player seeks to 01:00 minute mark and displays `Replay` icon in-place of `Play/Pause` button in the control-bar. This 01:00 minute mark is the duration of the first item in the first Canvas.

Similar behavior can be observed for `Annotations` component as well.

### Done Looks Like
Transcripts/Annotations components,
- [ ] seek the player's playhead to the correct time point when clicked on a cue/annotation for a multi-item Canvas
- [ ] can auto-scroll the transcript/annotations highlighting the correct cue/annotation while the player is playing the media jumping from item to item within the Canvas

joncameron:
Working great 🎉

PR #889, opened 2025-12-04
Fix transcript/annotation sync with playback for multi-source Canvas

Dananji:
Related issue: #884 

Changes in this PR:
- Adjust time calculations for transcript cue/annotation synchronization with playback time to consider alternate start times when there are multiple sources
- Adjust `currentTime` calcualation in the `useMediaPlayer` custom hook to consider alternate start times for multi-sourced canvases
- Refactor the playback time synchronization for transcript cues/annotations to remove duplicated code for time calculations

### DIFF
--- a/src/services/ramp-hooks.js
+++ b/src/services/ramp-hooks.js
@@ -580,9 +580,9 @@ export const useVideoJSPlayer = ({
     //if found set selected attribute on matching source then remove from currently marked one
     const originalQuality = sources?.find((source) => source.selected == true);
     const selectedQuality = sources?.find((source) => source.label == startQuality);
-    if (selectedQuality) {
-      originalQuality.selected = false;
+    if (selectedQuality && originalQuality) {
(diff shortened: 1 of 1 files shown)


cjcolvar (commented):

I was hoping to do some manual testing with the [IA manifest mentioned in the issue](https://iiif.archive.org/iiif/CSPAN3_20180217_164800_Poplar_Forest_Archaeology/manifest.json) but I'm getting a 504 when attempting to fetch it.

Also I tried the [opera one canvas cookbook manifest](https://iiif.io/api/cookbook/recipe/0064-opera-one-canvas/manifest.json) and it breaks when trying to go to the second source of the canvas.  This is the error I'm seeing:
```
TypeError: can't access property "selected", originalQuality is undefined
    setSelectedQuality ramp-hooks.js:584
    useVideoJSPlayer ramp-hooks.js:381
```
Is it expected to work or does that recipe's manifest need fixing?

Issue #872, opened 2025-10-14
Use Alternative Quality Level if Current is Unavailable


joncameron:
## Description
When a derivative for the current quality level is unavailable, Video.js displays an error message. In cases where the current quality level is unavailable (500 response from the m3u8 route or other error indicator), the player should attempt to try alternate streams. Currently users see the error message and it's not intuitive for them to try selecting a different quality level to resolve the issue. This may also only affect Wowza implementations.

Ordering of quality levels is set in the manifest.

## Quality level rotation
- If no quality is saved in localstorage:
  - Cycle through quality levels in sequence using the array defined in the manifest
- If quality is saved:
  - Try that quality, if there's an error, cycle through quality levels in sequence using the array defined in the manifest

## Done Looks Like
- [ ] Ramp tries other available streams if the currently selected stream has an error/is unavailable

joncameron:
Reading from localstorage at the moment; if the saved quality isn't present, it shows the error.

PR #890, opened 2025-12-05
Add fallback strategy for source selection

Dananji:
Related issue: #872 

Changes in this PR:

Implemented source fallback functionality for `VideoJS`,
-  for source selection with the `@silvermine/videojs-quality-selector` plugin. With this fix, when a source choice in quality selector fails to load, the player automatically cycles through available sources until one works. While trying available source choices it marks unavailable sources choices in the quality selector menu. If all sources fail an error message is displayed using `VideoJS`' error modal.
- for a multi-sourced Canvas where multiple segments are stitched together to build a single Canvas. With this fix; when a source fails to load while user is scrubbing/using structured navigation actions, the next viable available segment is loaded into the player. And the failed segment is marked as blocked in the progress-bar for the player session.

When the player encounters a failing source that particular source gets disabled in the quality selector for the player session (until page is reloaded, or switched to another Canvas) and an error is logged to the `devTools` console.

Screenshot 2025-12-05 at 3 37 14 PM

When all sources fail an error is displayed using `VideoJS`'s error modal.

Screenshot 2025-12-05 at 3 36 31 PM

When 2 consecutive source segments are failing in a multi-sourced Canvas the error message is displayed as a non-dismissable toast notification before loading the next viable available segment.

https://github.com/user-attachments/assets/8d8080dc-e7a4-442b-a068-6ce71ac3190b

### DIFF
--- a/src/services/transcript-parser.test.js
+++ b/src/services/transcript-parser.test.js
@@ -195,6 +195,13 @@ describe('transcript-parser', () => {
     });
 
     test('when transcripts list is not empty', async () => {
+      // Mock fetch to prevent actual network calls
+      const fetchSpy = jest.spyOn(global, 'fetch').mockResolvedValue({
+        status: 404,
+        headers: { get: jest.fn(() => 'text/plain') },
+        text: jest.fn(() => ''),
+      });
+
       const transcripts = await transcriptParser.sanitizeTranscripts(
         [
           {
@@ -235,6 +242,8 @@ describe('transcript-parser', () => {
           }
         ]
       });
+
+      fetchSpy.mockRestore();
     });
 
     describe('parses annotations from a Manifest URL in transcripts list', () => {


src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
```diff
@@ -305,6 +322,96 @@ function VideoJSPlayer({
...
       }
+
+      // Stop retrying if the player is currently in a fallback attempt
+      if (player.isFallingBack) {
+        e.stopPropagation();
+        return;
+      }
+
+      // Determine if this is a multi-source Canvas failure or source quality choice failure
+      /** ASSUMPTION: A Canvas is either multi-sourced OR multi-choice and not both at the same time */
```
cjcolvar: I assume this assumption is something we'll have to come back to when a use case requires it so thanks for documenting it!

cjcolvar (approved):
This code looks good to me.  I think it will be good to get this into Avalon to test it out thoroughly.

Issue #878, opened 2025-11-06
Handle VideoJS errors with unsupported src URLs gracefully


Dananji:
### Description

When the [video playlist Manifest](https://iiif.archive.org/iiif/taboca_201002_03/manifest.json) from Internet Archives is loaded into Ramp it shows a "Media could not be loaded. Network error or media format is not supported." error in the player and `"MEDIA_ERR_SRC_NOT_SUPPORTED: The media resource indicated by the src attribute was not suitable."` in the console in DevTools.
This error blocks the user from accessing the other canvases in the Manifest using previous/next buttons because the player's control-bar is hidden with error display inside the player. And this particular Manifest doesn't have a `structures` property for Ramp to build structured navigation below the player, which provides an alternate way to access other canvases in the Manifest.

### Done Looks Like
- [ ] When an unsupported source URL is loaded into the player, display the control-bar so that the user can navigate between canvases if the Manifest has multiple canvases and switch the current source if there are multiple sources for the current Canvas

PR #893, opened 2025-12-10
Display player controls in error state

Dananji:
Related issue: #878 

Changes in this PR:
- display player controls in the player's error state for multi-canvas manifests to provide users the ability to navigate between canvases. In this state only the previous/next buttons are enabled to allow canvas navigation
multi-canvas error

- hide player controls in the player's error state for single-canvas manifests
single-canvas error


cjcolvar (approved):
Looks good as far as I can tell without testing.  Nice to have this better UX!

PR #922, opened 2026-03-26
Create a standalone bundle for non-React apps

Dananji:
Related issue: #920 

Changes in this PR:
- adds `preact` as a `devDependency`
- adds additional config to the `vite.config.js` to create the standalone bundle with `preact/comact`
- document the usage of the standalone bundle in `README.md` (the unpkg URLs are the expected values for when/if this is published to NPM)


cjcolvar (approved):
Looks great!

PR #926, opened 2026-04-01
Save playback position

Dananji:
Related issue: #919 

Changes in this PR:
- adds a `resumeCache` prop to `MediaPlayer` with a default of `{ enable: false, ttlDays: 30, maxItems: 200 }` to configure the feature and LRU cache capacity in `localStorage`
- adds `usePlaybackPositions` custom hook implementing a bounded LRU cache in `localStorage` to persist per-canvas playback positions. This is in its own file `src/services/save-playback-positions.js` to keep the LRU implementation with the hook implementation.
- saves the current playback position on `timeupdate` events, skipping the first and last 5 seconds to avoid unhelpful resumes
- clears the saved position when the user reaches the end of the media
- shows a resume modal on initial page load if a saved position exists for the starting Canvas; modal is suppressed on canvas switches and when `startCanvasTime` is set in the props
- when the `resumeCache.enable` is set to `false` any existing saved playback positions in `localStorage` is cleared otherwise the browser keeps this unused information indefinitely with the feature disabled

Implementation notes for the resume modal: 
- its creation and display is invoked inside a `loadedmetadata` event, which is registered via `player.one()` method to listen for this event exactly once. This ensures the modal is shown only on the first time the player is loaded and suppressed for subsequent Canvas switches.
- resume modal is suppressed when `startCanvasTime` is set, as the explicit start time takes precedence
- a saved position is cleared immediately when the user takes action to resume playback/start from the beginning from the resume modal
- resume playback modal is not invoked in playlist context

The resume modal looks as follows for audio and video players;

Screenshot 2026-04-01 at 3 09 22 PM

Screenshot 2026-04-01 at 3 09 10 PM

Possible future work:
- show player controls along with the modal for easy navigation between canvases (just the previous/next buttons)
- re-organize the code structure to separate the custom hook files into a different directory for a cleaner code


cjcolvar (commented):
This is really neat and I'm excited to have this feature in Avalon!

This looks great for single canvas manifests but it seems like it isn't resuming playback when the save position is on a canvas other than the first.  For example, if I listen to https://media.dlib.indiana.edu/media_objects/tm70nc75d/manifest.json, which has two audio files, and stop playback at 20 seconds into the 2nd section and reload ramp with the manifest then ramp starts at 0 on the first canvas.  I'd expect ramp to know that the saved playback position for the manifest is on the 2nd canvas at 20sec and offer to resume me there.  This would be especially useful in a many sectioned item like https://media.dlib.indiana.edu/media_objects/nk322d54j.

Seeing that the local storage holds a canvas id -> playback position map, I'm wondering if this would need to be manifest id -> canvas id + playback position or if it is helpful to have two maps manifest id -> canvas id and canvas id -> playback position.  It is theoretically possible for a canvas to exist in more than one manifest and it seems like it could be helpful to resume playback regardless of the manifest.  But that is probably  too much of an edge case to make the extra complexity worth it.

Does this make sense and seem feasible?

cjcolvar (approved):
This looks great!  I'm excited to see this make it to production!

Not within scope of this PR, but I had a couple thoughts that I wonder might be future improvements if user testing agrees:
- For video items, seek to resume time and make modal semi-transparent to allow for some visual help when answering the question do I want to resume here
- For cross canvas resuming, change modal text to include section name to help with disambiguation instead of relying on noticing that the section title is bold (e.g. with https://media.dlib.indiana.edu/media_objects/tm70nc75d/manifest.json use something like "Resume playback from Lacquer Disc 1/1 Part 2 (40000002813915) at 00:31?").  It might be too much though.

Dananji:
New ticket to address the suggestions: https://github.com/samvera-labs/ramp/issues/932

Issue #924, opened 2026-03-30
[BUG] Duplicated caption option for multi-sourced Canvas in the player


Dananji:
### Description

The [IA Manifest](https://iiif.archive.org/iiif/CSPAN3_20180217_164800_Poplar_Forest_Archaeology/manifest.json) is a single Canvas Manifest with multiple items in the same Canvas. And it has a single `supplementing` Annotation for the Canvas that Ramp will render as a captions option in the player and a transcript.
Initially there is one captions option named 'auto' in the player. But, when the user switches between Canvas segments in the player (either via scrubbing or playback) the captions option gets duplicated.

### Done Looks Like
- [ ] Player doesn't duplicate caption options when switching between Canvas segments

joncameron:
👍

PR #933, opened 2026-04-09
Avoid text-track duplication for multi-source canvases

Dananji:
Related issue: #924

Changes in this PR:
- adds `srcIndexChangedRef` flag to track whether `srcIndex` has changed since the last player load
- uses `srcIndexChangedRef` to distinguish between an `emptied` event triggered by a source change on a multi-source Canvas versus a `qualityRequested` event from the quality plugin to prevent text-tracks from being duplicated on source change

non-duplicated-captions


cjcolvar (approved):
Looks good!

PR #937, opened 2026-04-16
Change popup menu direction based on player orientation

Dananji:
Related issue: #931 

Changes in this PR:
- Calculate the tallest popup menu and compare it against the clearance on top of the player to check whether there is enough space to display the popup menu without getting clipped by the viewport. Depending on this change the direction of the popup menu as needed
- Use `window.resize` and `window.scroll` events to re-calculate this metrics to change the direction of the popup menus as needed


cjcolvar (approved):
Neat!  This code looks good as far as I can tell.  I think we'll want to test it in avalon to make sure the show/hide of the menus isn't a problem. 🤞

PR #938, opened 2026-04-16
Fix captions icon disappear on canvas change

Dananji:
Fix a bug with rendering captions icon introduced in https://github.com/samvera-labs/ramp/pull/916

### DIFF
--- a/src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
+++ b/src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
@@ -739,11 +739,12 @@ function VideoJSPlayer({
 
     /*
       Update player control bar for;
-       - track scrubber button
-       - volume panel
-       - appearance of the player: big play button and aspect ratio of the player
+        - track scrubber button
+        - volume panel
+        - if tracks exists: captions button for video players
+        - appearance of the player: big play button and aspect ratio of the player
         based on media type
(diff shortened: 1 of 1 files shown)


src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
```diff
@@ -784,6 +785,14 @@ function VideoJSPlayer({
         player.trigger('volumechange');
       }
 
+      if (tracks?.length > 0 && isVideo && !controlBar.getChild('subsCapsButton')) {
+        let subsCapBtn = controlBar.addChild(
+          'subsCapsButton', {}, volumeIndex + 1
+        );
+        // Add CSS to mark captions-on
+        subsCapBtn.children_[0].addClass('captions-on');
```
cjcolvar: Are the captions always enabled here or does there need to be some logic to check?

Dananji:
Good catch!! I added a check for `startCaptioned` sticky setting for turning on captions.

Additionally, I added a fix to remove the stale `textTrack.change` event handlers between Canvas changes. The previous implementation was stacking on `textTrack.change` event handlers for each Canvas change, which led them to read stale values in `localStorage` to cause unexpected behavior around forced text-tracks.

cjcolvar (approved):

Issue #714, opened 2024-11-04
[BUG]Playback fast-forwards or pauses unexpectedly when switching tabs on muted media


abankhele:
**Describe the bug**
On iPad, if the audio or video player is muted and the page is reloaded, switching to another tab causes the playback duration to fast-forward beyond the time spent away. On desktop and Android, playback pauses when the tab is inactive and resumes from the previous timestamp upon returning.

**To Reproduce**
Go to 'https://avalon-dev.dlib.indiana.edu/media_objects/x920fw86z'  ensure it is muted, and click Play.
Switch to another tab and remain there for a set time (e.g., 10 seconds).
Return to the media tab.
Observe:
On iPad, the playback duration has advanced beyond the time spent on the other tab.
On desktop and Android, playback pauses while away, then resumes from the paused timestamp upon returning.

**Expected behavior**
Playback should continue smoothly from the correct duration on all devices, without pausing or fast-forwarding when switching tabs.

**Screenshots**
Attached video are for ipad and desktop 
[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdzZVQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--3fa00ceca47b225e41045d6411e6cde4dac6558d/Screen%20Recording%202024-11-04%20at%203.35.08%E2%80%AFPM.mov)

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeEtVQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f9c362e904900cdbffcb7afe74d984b1d6008c7f/Screen%20Recording%202024-11-04%20at%203.36.14%E2%80%AFPM.mov)

**Desktop (please complete the following information):**
 - OS: macos
 - Browser: Chrome

**Smartphone (please complete the following information):**
 - Device: ipad, andriod
 - OS: ios 16, andriod 13
 - Browser: safari, chrome

joncameron:
This should be checked against a non-Avalon page; time didn't seem to continue while Ramp was in the background. If both tabs are Avalon pages with Ramp running, this overlap could be the cause.

Dananji:
[...]

PR #939, opened 2026-04-17
Fix active and inactive tab playback and stalling for player

Dananji:
Related issue: #714

Changes in this PR:
- adds a `visibilitychange` listener in `initializeEventHandlers` (`ramp-hooks.js`) that pauses the player when the tab becomes inactive and resumes it when the tab becomes active again, preventing the playhead from advancing while the tab is inactive on iOS/iPadOS
- on iOS, replaces `Math.max(currentTimeRef.current, player.currentTime())` with `currentTimeRef.current` when setting the seek target in `playerLoadedMetadata` event, because iOS can advance `player.currentTime()` in the background before the `visibilitychange` pause takes effect, causing the player's `seek` event to jump-forward on next load
- adds `stalled` event listener for iOS in to recover from stalling that happens in audio player. When VideoJS `stalled` event fires with `readyState < 3`, call `player.load()` and `player.play()` to un-stuck the player

### DIFF
--- a/src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
+++ b/src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
@@ -290,6 +290,28 @@ function VideoJSPlayer({
       setStartVolume(player.volume());
     });
     /**
+     * On iOS, the OS audio session can be revoked by another tab or app, leaving the
+     * player in a stuck state where the playhead moves but no audio is produced, or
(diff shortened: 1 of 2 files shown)


cjcolvar (approved):
Great job handling these iOS quirks!  This looks good to me but there are some merge conflicts that need to be resolved.

cjcolvar (approved):

Issue #941, opened 2026-04-22
Ramp release for Avalon 8.2


Dananji:
### Description

Create a Ramp release (5.1.0) to be included in Avalon's 8.2 release. This would include the following new features;
- AD feature (https://github.com/samvera-labs/ramp/issues/839, https://github.com/samvera-labs/ramp/issues/850)
- Resume playback (https://github.com/samvera-labs/ramp/issues/919)
- Standalone build of Ram for non-React environments (https://github.com/samvera-labs/ramp/issues/920)

And bug fixes;
- When media cannot be played, error message in media player should be keyboard accessible https://github.com/samvera-labs/ramp/issues/660
- [BUG]Playback fast-forwards or pauses unexpectedly when switching tabs on muted media (iPad) https://github.com/samvera-labs/ramp/issues/714
- [BUG] Duplicated caption option for multi-sourced Canvas in the player https://github.com/samvera-labs/ramp/issues/924
- [BUG] Dimension-based popup menu direction for audio players https://github.com/samvera-labs/ramp/issues/931

### Done Looks Like
- [x] Dependency updates for older versions in the dev and prod dependencies
- [ ] Update code in Netlify demo site and Styleguidist docs in GitHub pages
- [ ] Readme and Wiki documentation created/updated as needed
- [ ] Update version numbers in package.json
- [ ] Create and publish tag for next release
- [ ] Create release in GitHub for next release
- [ ] Publish release to NPM

joncameron:
Ready for release and tag + publishing through NPM.

PR #942, opened 2026-04-22
Dependency audit and upgrade as needed

Dananji:
Related issue: #941 

Changes in this PR:
- remove `react-compiler-runtime` from dev dependencies, since the compiler runtime is built into React 19
- update `@types/react` from the fixed `18.3.5` to `^19.0.0` to match React 19
- upgrade `@testing-library/jest-dom` from v5 to v6
  - update `jest.setup.js` to import `@testing-library/jest-dom` directly, because the `extend-expect` sub-path was removed in v6
- upgrade `jest`, `babel-jest`, and `@types/jest` from v26 (EOL) to v29
  In Jest 26, `jsdom` was the built-in default test environment and it simulated a browser DOM automatically, so `window`, `document`, etc. in unit tests were always available without any configuration. But, from `jest` 27;
  - the default `testEnvironment` was switched from `jsdom` to `node` - meaning tests run in a plain Node.js environment with no DOM by default, which requires setting `testEnvironment: 'jest-environment-jsdom'` in `jest.config.js`
  - `jest-environment-jsdom` was extracted into a separate package, so it must be explicitly installed


cjcolvar (approved):
Yay to updated dependencies! 🎉

PR #947, opened 2026-05-07
Cdl playlist items

Dananji:
Related issue: https://github.com/avalonmediasystem/avalon/issues/5001

Changes in this PR:
- add `VideoJSErrorModal.js` with a `showErrorModal` function to display a custom uncloseable error modal when media is inaccessible (e.g. CDL-gated playlist items)
- use VideoJS's config `vhs.maxPlaylistRetries: 0` to prevent VHS from retrying CORS-blocked/403 HLS manifests indefinitely. Then using `retryplaylist` event from VHS to detect the exhausted HLS playlist retries and display an error modal to the user. In audio-only mode the error message is compressed into a single line to avoid overflow. For multi-canvased manifests, previous/next buttons are left active while the other controls are blocked so that the user can navigate between sections/playlist items.
- defer the resume modal to `loadedmetadata` event both same-Canvas and cross-Canvas playback resumes so it does not stack on top of an existing error modal when media fails to load

Multi-source Canvas trying each segment until it reaches the end and displays the error message;

https://github.com/user-attachments/assets/24384774-e135-426e-a24c-bb7703748c47

Multi-source unavailable message display in an audio player;
audio-multi-source-unavailable-message

Multi-canvas source unavailable message display in an audio player with previous/next buttons enabled;
audio-player-unavailable-message

### DIFF
--- a/src/components/MediaPlayer/MediaPlayer.js
+++ b/src/components/MediaPlayer/MediaPlayer.js
@@ -110,7 +110,11 @@ const MediaPlayer = ({
       // where the native iOS player is used by default or on fullscreen-mode.
       // i.e. Both Safari and Chrome on iPhones, only Chrome on iPads.
       html5: {
-        nativeTextTracks: !IS_ANDROID && ((IS_IPAD && !IS_SAFARI) || IS_IPHONE)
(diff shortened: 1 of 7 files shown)


cjcolvar (approved):
Looks good as far as I can tell.  With the amount of changes in this PR I suspect we want to hold it for a ramp release that isn't going in Avalon 8.2.

PR #949, opened 2026-05-08
Scope playback pause on inactive tab to iOS

Dananji:
This PR fixes a regression introduced in #939 where the `visibilitychange` handler was registered unconditionally, causing the player to pause and resume on tab switch for all browsers and platforms. This`visibilitychange` handler should only be applied in iPads.


cjcolvar (approved):

Issue #946, opened 2026-05-04
Captions Overlap When Navigating with Transcript Component


joncameron:
## Description
In some cases caption text is not cleared when users select a new cue in the Transcript component, leading to visual artifacting.

<img width="941" height="565" alt="Image" src="https://github.com/user-attachments/assets/6b86adbb-247e-4afb-b989-e4e0f3589ff9" />

## To Reproduce
- Load https://iiif.archive.org/iiif/CSPAN3_20180217_164800_Poplar_Forest_Archaeology/manifest.json in the Ramp demo site
- Play, then pause the video
- Select multiple timestamps close in time to navigate the playhead to those cues
- Previous caption text is not cleared when the new caption text is drawn, leading to visual overlap
- This can also be observed during playback, where navigating between cues causes a visual flash due to the existing caption being rendered right up until the point the new caption is rendered

PR #951, opened 2026-05-11
Fix captions overlap with transcript cue clicks

Dananji:
Related issue: #946

When there are adjacent transcript cues that share an end time and a start time (a common boundary), both cues get highlighted and and displayed in the VideoJS player when seeking. Adding a 1ms offset to the playback time when synchronizing the transcript component with the VideoJS player fixes this.

The screen recording shows the transcript cue highlighting and VideoJS caption display updates when player is paused for the following 3 cues in its transcript,
```
37
00:01:20.000 --> 00:01:24.000
LANDSCAPE,DERSTAND

38
00:01:24.000 --> 00:01:28.000
THE HOUSES HE DESIGNED.
WE ARE ALSO USING IT AS THE

39
00:01:28.000 --> 00:01:31.000
LIVES OF ENSLAVED PEOPLE.

```

https://github.com/user-attachments/assets/1286ba30-08bd-4fc1-8be5-ffda59bea651

### DIFF
--- a/src/components/Transcript/Transcript.js
+++ b/src/components/Transcript/Transcript.js
@@ -489,7 +489,11 @@ const Transcript = ({ playerID, manifestUrl, showMetadata = false, showNotes = f
 
   const seekPlayer = useCallback((time) => {
     setCurrentTime(time); // so selecting an item works in tests
-    syncPlayback(time);
+    /* Add 1ms offset so that, the player time lands strictly inside the current cue.
+    When 2 adjacent cues share a boundary, i.e. end time of cue 1 === start time of cue 2,
+    VideoJS treats both as active at the boundary time, which shows both captions
+    over the other. */
+    syncPlayback(Math.round((time + 0.001) * 1000) / 1000);
   }, []);
 
   if (!isLoading) {
--- a/src/components/Transcript/Transcript.test.js
+++ b/src/components/Transcript/Transcript.test.js
@@ -124,7 +124,7 @@ describe('Transcript component', () => {
         // Click on the cue's timestamp
         fireEvent.click(transcriptItem.children[0]);
         expect(transcriptItem.classList.contains('active')).toBeTruthy();
-        expect(syncPlaybackMock).toHaveBeenCalledWith(1.2);
+        expect(syncPlaybackMock).toHaveBeenCalledWith(1.201);
       });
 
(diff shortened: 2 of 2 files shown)


cjcolvar (commented):
This is a clever workaround.

Does this cause any problems with the beginning or end of the stream?  
I want to make sure we don't introduce an off by 1 issue since seeking won't be to exactly 0 and seeking to the end will be just past the end.  Alternatively you could probably add those two base cases here that makes 0 be 0 and the end be the end.  Thoughts?

Dananji:
Thank you for pointing that out! It's a very valid point that I hadn't fully considered :)

The new offset would indeed be an issue for a cue with a duration less than 1ms. For instance, a cue with a 1ms duration -- such as `{ start: '00:00:00.000', end: '00:00:00.001' }` would be resolved to `seek(0.001)` using the start time with this adjustment. This would still keep the player's current time within the cue's boundaries.

As you noted, this offset might cause the player to jump to the next cue instead of displaying the current cue, but only for cues with a duration < 1ms. According to the [VTT spec](https://www.w3.org/TR/webvtt1/#file-structure), a VTT timestamp has a precision of thousandths of a second (1ms). I also tested a sample VTT in the [W3C VTT validator](https://w3c.github.io/webvtt.js/parser.html) which flagged this issue in the timestamps.

Because of this, if such a cue were to exist in a given VTT file, it would likely be considered an invalid cue, which VideoJS would ignore regardless of this offset. Therefore, it seems unlikely to introduce a regression in practice.
So, introducing the two base cases might add a bit of complexity for a scenario that may not occur in the real-word usage, as cues shorter than 1ms are considered invalid.

Let me know what you think!

cjcolvar:
Sorry, I wasn't actually thinking about short cues but that makes sense.  I was wondering if this change has any affect on other parts of ramp around handling the beginning or end of a file/canvas.  Maybe I'm reading things wrong but it seems like `syncPlayback` will actually change the player current time so I was afraid of any code or hooks that expected exact 0 of the beginning of a file or the exact ending duration of the file/canvas/clip wouldn't get triggered as expected since the time would be shifted by 1ms.  I did some quick testing and wasn't able to break anything so maybe the code is resilient to this small fuzziness.

cjcolvar (approved):

Issue #953, opened 2026-05-13
[BUG] Switching to unavailable source quality shows an error


Dananji:
### Description

Recently some work was done around cycling through quality sources when a selected quality is not available in Ramp player (related issue: https://github.com/samvera-labs/ramp/issues/872).
With this work Ramp cycles through the qualities and selects an available quality or display an error message if it fails. 
When an unavailable quality source is selected during playback the player skips the unavailable source and goes back to the working source, however it doesn't restore to the previous player time and continue with the playback.

It seems the player is running into an error while trying to seek to the previous playback time after the switch and continue playback.

### Done Looks Like
- [ ] Ramp selects a working source when an unavailable source is selected
- [ ] Restore the player's state once the switch is made, i.e. seek to the previous time and resume playing if the player was playing at the time of the switch

PR #954, opened 2026-05-15
Restore the playback state in the source fallback logic

Dananji:
Related issue: #953

Changes in this PR:
- memoize the playback state via `wasPlayingRef` in `qualityRequested` event before the source changes and use it to restore playback state after the fallback logic resolves the source selection to a working source
- use `player.one('loadedmetata', ..)` event instead of `player.ready()` to reset the `isFallingBack`, because this makes sure a working source is loaded into the player unlike `player.ready` which fire synchronously if the player is already ready
- use `player.one('seeked', ...)` event emitted by the quality-selector plugin after loading a valid source in the fallback logic to restore `currentTime` and resume playback with the memoized values. This ensures, the quality-selector's `seeked` event doesn't override the state reset on the Ramp side.
- add unit tests for the playback state restoring in the source fallback logic

### DIFF
--- a/src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
+++ b/src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
@@ -168,6 +168,9 @@ function VideoJSPlayer({
   // Flag to track whether srcIndex has changed since last load
   const srcIndexChangedRef = useRef();
 
+  // Flag to track whether the player was playing during a fallback for source selection
+  const wasPlayingRef = useRef();
+
   /**
    * Setup player with player-related information parsed from the IIIF
    * Manifest Canvas. This gets called on both initial page load and each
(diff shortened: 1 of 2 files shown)


src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
```diff
@@ -326,6 +329,11 @@ function VideoJSPlayer({
       handleTimeUpdate();
     });
     player.on('qualityRequested', (e, quality) => {
+      /* Remember the player state before source selection changes, which is used in the error
+      handler to determine whether to resume playback after fallback efforts in source selection. */
+      if (wasPlayingRef.current == undefined) {
```
cjcolvar, line 339: I assume this guard is to avoid setting the wasPlayingRef multiple times and possibly overwriting the original value if the quality selector has to fallback.  Is this ref cleared (set back to undefined) anywhere so this can get set another time?  Or does this ref get rebuilt on player load/reload so it doesn't actually need to be cleared?

Dananji: Good catch!
Yes, the guard is there to prevent overwriting the original playback state during a fallback chain, since the fallback itself triggers another `qualityRequested` event.

You're right that the ref is never cleared after the fallback completes. I think we can safely reset the `wasPlayingRef.current` back to `undefined` at the end of the `player.one('seeked', ...)` callback, which is the successful completion point of the fallback sequence.

cjcolvar (approved):
👍

PR #955, opened 2026-05-21
Fix styling in the form in demo site

Dananji:
Some minor styling changes to the form in the demo site.

**Before:**
before

**After:**
after


cjcolvar (approved):
👍

Issue #320, opened 2023-12-14
IIIF Authorization Flow 2.0 Support


jcoyne:
**Is your feature request related to a problem? Please describe.**
We have videos that require authorization prior to viewing.

**Describe the solution you'd like**
Implement IIIF Auth V2 https://ramp.avalonmediasystem.org/

**Describe alternatives you've considered**
Build a new viewer.

## Auth Flow Implementation
Basic Authorization flow:

- Ramp requests Access Service endpoint, appending an origin value. This should open a new tab/window.
- Once action has been taken in the new tab, Ramp should detect the tab closing and then request the Token Service endpoint. This is done within a hidden iframe. Ramp will need an event listener to receive cross domain requests.
- Receive a postMessage containing the JSON response. Ramp will need to have error handling for bad requests and failed authentication. Successful responses will contain the access token needed for the probe service.
- Request the Probe service, with the access token included in an authorization header Authorization: Bearer .... This should be a GET request and expect a JSON response. Access errors should be handled, success should render the content.

## Done Looks Like
- [ ] Ramp is able to parse the new auth structure: Probe → Access Service → [Token, Logout]
- [ ] When a IIIF Auth Flow 2.0 service is defined in a manifest, Ramp works through the auth flow detailed above

elynema:
From Justin Coyne at Stanford:

Here's one https://purl.stanford.edu/ks701xw3226/iiif3/manifest Some code is in here: https://github.com/sul-dlss/purl/blob/main/app/models/iiif3_presentation_manifest.rb

masaball:
Notes for implementation:
* Ramp will need to parse the new auth structure. Probe -> Access Service -> [Token, Logout]
* Basic Authorization flow is laid out [here](https://iiif.io/api/auth/2.0/#simple-authorization-flow):
    * Ramp requests Access Service endpoint, appending an [origin value](https://iiif.io/api/auth/2.0/#interaction-with-access-services). This should open a new tab/window.
    * Once action has been taken in the new tab, Ramp should detect the tab closing and then request the Token Service endpoint. This is done within a [hidden iframe](https://iiif.io/api/auth/2.0/#access-token-service-request). Ramp will need an event listener to receive cross domain requests. 
    * Receive a postMessage containing the JSON response. Ramp will need to have error handling for bad requests and failed authentication. Successful responses will contain the access token needed for the probe service. 
[...]

PR #959, opened 2026-06-01
Iiif auth

Dananji:
Related issue: #320

Changes in this PR:
- read and parse IIIF Auth 2.0 related services from a Canvas Annotation `body` including support for `Choice` Annotation `body`
- display an overlay with login action implemented as a separate component called `AuthOverlay` that renders over the player when a Canvas has protected behind a IIIF Auth 2.0 service
  - displays all available text information from the auth service as needed
  - if a `heading` text exists display it at the top of the login UI (e.g. 'Please Log In' in Avalon Manifest)
  - display `label` text below the heading in bold text
  - if a `note` text exists display it below the `label` text
- add a new module called `auth-service.js` to handle the probing, re-probing, access token acquisition via hidden iframe in the auth flow
- add state variables to keep track of the auth status of the current Manifest across components
- move sanitizing HTML using `DOMPurify` to `utility-helpers.js` as a util function called `sanitizeHtml`

Stanford IIIF Manifest with auth:
stanford-auth-manifest

Avalon IIIF Manifest with auth and login waiting:
avalon-auth-manifest
avalon-auth-manifest-waiting

### DIFF
--- a/src/components/MediaPlayer/VideoJS/AuthOverlay.js
+++ b/src/components/MediaPlayer/VideoJS/AuthOverlay.js
@@ -1,0 +1,219 @@
+import React, { useEffect, useRef, useState } from 'react';
+import PropTypes from 'prop-types';
+import { getLabelValue, sanitizeHTML } from '@Services/utility-helpers';
+import { probeResource, requestTokenViaIframe } from '@Services/auth-service';
+import './VideoJSPlayer.scss';
+
+/**
+ * Display authentication dialog and handle login/logout interactions for IIIF resources
+ * protected by IIIF Auth 2.0 spec. 
+ * It is rendered over the VideoJS player when a Canvas has an AuthProbeService2 defined.
+ * The flow is as follows;
+ * 1. On page load: probe the resource without a token to determine if login is required
+ * 2. Login: performs the authnentication flow based on the access service profile and gets a token
+ * 3. On token received: re-probe with token to confirm authorization, or show error if probe fails
+ * 4. Logout button (if present): open logout URL in new tab, clear token
+ * @param {Object} props
+ * @param {Object} props.authService
+ * @param {String} props.authToken
+ * @param {String} props.authStatus
+ * @param {Function} props.onTokenReceived
+ * @param {Function} props.onAuthStatus
+ */
+function AuthOverlay({ authService, authToken, authStatus, onTokenReceived, onAuthStatus }) {
+  const [errorMessage, setErrorMessage] = useState(null);
+  const [isLoggingIn, setIsLoggingIn] = useState(false);
+  const loginTabRef = useRef(null);
+  const pollIntervalRef = useRef(null);
+
+  const { probe, accessService, tokenService, logoutService } = authService;
+
+  // Probe without token on page load to determine if login is required
+  useEffect(() => {
+    if (authStatus !== 'idle') return;
+    onAuthStatus('probing');
+    probeResource(probe.id, null)
+      .then(({ status }) => {
+        if (status === 200) {
+          onAuthStatus('authorized');
+        } else {
+          onAuthStatus('login-required');
+        }
+      })
+      .catch(() => {
+        onAuthStatus('login-required');
+      });
+  }, []);
+
+  // Cleanup login tab pollin interval on unmount
+  useEffect(() => {
+    return () => {
+      clearInterval(pollIntervalRef.current);
+    };
+  }, []);
+
+  /**
+   * Acquire token via a hidden iframe using postMessage API, then save this token
+   * and update the auth status in state.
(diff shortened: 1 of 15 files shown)


src/services/iiif-parser.js
```diff
@@ -715,3 +710,52 @@ export function getSearchService(resource) {
...
+    const serviceArray = Array.isArray(authService) ? authService : [authService];
+    const services = serviceArray.flatMap(s => Array.isArray(s) ? s : [s]);
+
+    // Find and parse service type="AuthProbeService2"
+    const probe = services.find(s => s.type === 'AuthProbeService2');
+    if (probe) {
+      // Read probe services as an array
+      const probeServices = Array.isArray(probe.service)
+        ? probe.service
+        : probe.service ? [probe.service] : [];
```
cjcolvar: I think it would be nice to unnest these ternary operators for readability.

src/components/MediaPlayer/VideoJS/AuthOverlay.js
```diff
@@ -0,0 +1,212 @@
...
+  };
+
+  /**
+   * Handle login button click based on the access service profile:
+   * - kiosk: acquire token via iframe
+   * - active: open a login tab, poll for it closing, then acquire token via iframe
+   * If popup is blocked, fall back to direct token acquisition without opening a login tab
+   */
+  const handleLogin = () => {
+    if (!accessService || isLoggingIn) return;
```
cjcolvar: Should `!accessService` set auth status to `error` and set an `errorMessage`?  

I'm thinking if the manifest is invalid and doesn't contain the access service and a user clicks the login button then would the flow get stuck with the button saying waiting?  Or is this scenario guarded against at a higher level than the overlay?

src/services/auth-service.js
```diff
@@ -0,0 +1,102 @@
...
+
+    const onMessage = (event) => {
+      /* Validate event.origin against token service origin to prevent accepting
+       messages from malicious origins */
+      if (event.origin !== expectedOrigin) {
+        reject(new Error('Invalid origin'));
+      }
+
+      const data = event.data;
+      if (!data || (data.messageId && data.messageId !== messageId)) return;
```
cjcolvar, line 61: Should this case return a reject error?

cjcolvar (commented):
This looks so good!  Love how clean it is as well as great comments and tests!

I had a few questions but this looks really close.

src/components/MediaPlayer/VideoJS/AuthOverlay.js
Dananji: Good catch 💯 You're right! The UX is broken at this stage.

The `!accessService` guard would prevent the button from getting stuck on `"Waiting…"` since `setIsLogginIn(true)` runs after it. But the user would be waiting indefinitely for nothing without any feedback because, this essentially is a no-op without a valid `accessService`.
And this UI is partially guarded at a higher level because `AuthOverlay` is only rendered when there's a valid `authService` which has `accessService` as an optional field.
To fix it, I think we can split the guard to first check `isLoggingIn` to return silently and then checking `!accessService` to display an error message. This would provide better UX.

src/services/auth-service.js
Dananji: Silently returning here is intentional because, any page script can fire a `postMessage` from the same origin with a different `messageId`.
Therefore, rejecting here could potentially abort a legit auth flow on the first unrelated message that happens to arrive from origin, however the `return` would keep the listener alive until a message that actually matches the given id arrives.

src/components/MediaPlayer/VideoJS/AuthOverlay.js
Dananji: I just re-read the IIIF Auth 2.0 spec, and I had missed the fact that the [probe service](https://iiif.io/api/auth/2.0/#probe-service-description) is _required_ to have a `service` prop with one or more access services with `type="AuthAccessService2"`. 

Each of these [access services](https://iiif.io/api/auth/2.0/#access-service-description) are then _required_ to have a `service` array with references to access token and other related services.  Out of these services there _must_ be exactly one access token service and _may_ contain logout services.

So, ideally Ramp shouldn't be displaying the `AuthOverlay` with a login button if there are no valid `accessService` and `tokenService` in the `authService` object parsed from the given probe service.

Dananji:
With this last change, the previous changes to guard against missing `accessService` in the login button click handler is not needed because the user won't see a login button when either `accessService` or `tokenService` is missing in the given probe service. Ramp will display the following UI for this use-case;

restricted-content

The text in this UI is read from the probe service's `errorHeading` and `errorNote` props if they exist. Otherwise a default message is displayed.

cjcolvar (approved):
Looks great!

PR #960, opened 2026-06-01
Fix empty playlist annotation bug

Dananji:
Guard against `null` `annotationSet` for empty playlists when there are zero canvases. Without this check, Ramp breaks and displays an error trying to filter through annotations from where there are `null` elements.


cjcolvar (approved):
Good catch!

PR #961, opened 2026-06-04
Setup Storybook

Dananji:
Related issue: #956 

Changes in this PR:
- initial setup of storybook library with necessary configurations
  - `.storybook/main.js` configures the framework, `@storybook/addon-docs` and `@storybook/addon-a11y` addons, story globs for `*.mdx` and `*.stories.jsx`, and Vite aliases for `@Components` and `@Services`
  - `.storybook/preview.js` imports `video.js` CSS globally so it is available across all stories and sets up the controls and editor in the docs UI
- replace the `Styleguidist` related scripts with relevant `Storybook` commands
- add docs for the `MediaPlayer` component with one story to establish the documentation pattern for the docs. This component was chosen because, this seemed to be the most complex with the VideoJS integration.
  - use `JSDoc` comments in `MediaPlayer.propTypes` entries to populate the auto-generated props table in `Storybook`
  - remove previous `.md` doc file for `MediaPlayer`
- add `storybook-static` (storybook built artifact) and `debug-storybook.log` to `.gitignore`
- use Vite's `import.meta.env.DEV` to detect the environment to load the VideoJS language map accordingly for docs site and demo site (this needs further testing for the prod environment), which was broken with the previous use of `process.env.NODE_ENV`

Issue #956, opened 2026-05-27
Swap documentation library in Ramp from Styleguidist


Dananji:
### Description

Implement the docs site with the selected documentation library from the investigation done in https://github.com/samvera-labs/ramp/issues/952

### Done Looks Like
- [x] Install new library and its required dependencies
- [x] Create/edit the docs files to populate the docs site
- [x] Cleanup the Styleguidist dependency and related dependencies
- [x] Cleanup `md` files related to the Styleguidist docs pages

joncameron:
5 points for Storybook; 3 points for Docusaurus.

joncameron:
Looks amazing!! This is so much nicer than the Styleguist docs.

PR #962, opened 2026-06-09
Add sample storybook manifests and clean up props and state in stories

Dananji:
Related issue: #956

This is the 2nd PR for this work. Other PRs: #961 

Changes in this PR:
- create a `hashArgs` util function to generate a hash for the given prop arguments. This value is used in `IIIFPlayer` component on storybook stories to create a unique key for each prop configuration, which forces React to re-render the component as the user changes the props using controls
- Extract common prop `manifestUrlControl` into `preview.js` as a shared storybook util. This prop setup in stories allow each story to extract `manifestUrl` in `IIIFPlayer` as a control in the panel
- Move `videoJSLanguageOptions` into `preview.js` as a shared storybook util, since this could be re-used in other stories
- setup storybook's own set of sample manifests to be used in the stories. These manifests use the resources setup in the repo. Stories use these manifests using the `IIIFPlayer`'s `manifestUrl` prop instead of using the `manifest` object
- add `manager.js` with `storybook/manager-api` to allow customize the control UI in stories
- create stories and `.mdx` files for `Annotations` component and remove its `.md` file


cjcolvar (approved):

PR #965, opened 2026-06-11
Add the rest of the docs and stories to Storybook

Dananji:
Related issue: #956 

This is the 4th PR for this work. Other PRs: #961, #962, #963

Changes in this PR:
- add stories to `StructuredNavigation`, `SupplementalFiles`,`Transcript` components
- fix `search` prop related conditional rendering which was a bug that was missed before

PR #963, opened 2026-06-09
Add stories to AutoAdvanceToggle, IIIFPlayer, MetadataDisplay components

Dananji:
Related issue: #956 

This is the 3rd PR for this work. Other PRs: #961, #962

Changes in this PR:
- add stories to `AutoAdvanceToggle`, `IIIFPlayer`, `MetadataDisplay` components
- add fallback logic for auto-advance feature in `VideoJSPlayer` for use-cases where `AutoAdvanceToggle` component is used without `StructuredNavigation`

PR #966, opened 2026-06-12
Remove Styleguidist and related files/configs

Dananji:
Related issue: #956 

This is the 5th PR for this work. Other PRs: #961, #962, #963, #965

Changes in this PR:
- remove `Styleguidist` dependencies and scripts from `package.json` and `yarn.lock`
- remove `Styleguidist` and its related,
  - build output files in `docs/**`
  - configuration in `styleguide.config.js`
  - Git workflow for deployment in`.github/workflows/docs-deploy.yml`
- update `CONTRIBUTING.md` (this will be added back in with `Storybook` related information in future work)
- add `esbuild.loader` config  to `vite.demo.config.js` for the Vite dependency scanner to use in the demo environment 
- add missing `html-entities` and `lodash` dependencies which were previously transitive dependencies pulled in indirectly via `webpack` and `react-styleguidist` (respectively) into the dependency tree


cjcolvar (approved):
🎉

Issue #957, opened 2026-05-27
Deploy new docs site to GitHub pages


Dananji:
### Description

The goal of this is to setup the GitHub workflow for the deployment of the new docs site to GitHub pages site.
Using GitHup Pages' source setting `"GitHub Action"` instead `"Deploy from a branch"` (the current setting with Styleguidist), the docs files can be removed from the `main` branch.

### Done Looks Like
- [x] Setup the docs deployment to the GitHub pages site via `actions/deploy-pages`
- [x] Site is updated when a docs change is made

PR #970, opened 2026-06-15
Add Github workflow to build and deploy Storybook to GitHub pages

Dananji:
Related issue: #957

Changes in this PR:
- add a GitHub Actions workflow `storybook-deploy.yml` to build and deploy the `Storybook` docs site to GitHub Pages
  - this workflow is invoked by pushes to `main` branch that has changes made to files in `src/**` or `.storybook/**`
- change `loadVideoJSLanguage` in `videojs-language-loader.js` to statically glob all VideoJS language JSON files via `import.meta.glob`, ensuring Vite bundles each language file as its own chunk for all deployments 
  - the previous fully dynamic `import()` with `@vite-ignore` was not picked up by the `Storybook` docs build, causing language change requests from `Storybook` prop controls to fail with 404 silently
  - this bumps the bundle size roughly ~150KB (for the 105 lazy-loaded per-language chunks each ~1KB) because of the language JSON files that are bundled and duplicated for some use-cases, but this provides a simple bundling option for all possible deploy environment without complicating the code and configs
- update `CONTRIBUTING.md` to document the new GitHub Pages docs deployment
- add `rimraf` as a dev dependency which was removed in #966 that existed as a transitive dependency  of `webpack-dev-server` 
- add `Storybook` prod manifests for the GitHub pages deployment
- remove `.storybook/manager.js`, which had config that is not needed for this phase


.github/workflows/storybook-deploy.yml
```diff
@@ -0,0 +1,52 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
```
cjcolvar: Is there a reason for using version 20 or can this be bumped to the latest tested with ramp?
```suggestion
          node-version: '24'
```

cjcolvar (approved):
Nice work with this!  I know you worked hard trying to figure out a dynamic loading for language data before but this seems fine and prioritizes i18n.  I don't think it is worth it at this point to spend a lot of time to try to save the 150K.

Issue #972, opened 2026-06-15
Add prop descriptions for Storybook prop controls


Dananji:
### Description

In the current implementation of `Storybook` docs, the props control panel for each story doesn't provide any description as to what each prop does when they are changed. While the naming of some props are straight-forward, there are some that can be confusing.

### Done Looks Like
- [x] Set the configuration needed to display prop descriptions in the prop table in stories
- [x] Add description for each prop control as needed

joncameron:
💯

PR #974, opened 2026-06-16
Enable prop descriptions in controls

Dananji:
Related issue: #972 

Changes in this PR:
- enable prop descriptions in the prop controls panel using `parameters: { controls: { ..., expanded: true } }` in `preview.js`
- override some of the prop descriptions populated from the `.js` file's prop descriptions to simplify the role of the prop
- add a new style sheet for styling components and others in the Storybook docs site as needed

Additionally, I enabled `code` tab in the control panel using `docs: { codePanel: true }` in `preview.js`

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,4 +1,5 @@
 import 'video.js/dist/video-js.css';
+import './storybook-styles.scss';
 
 /**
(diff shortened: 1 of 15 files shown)


cjcolvar (approved):
👍

Issue #958, opened 2026-05-29
IIIF Auth 2.0 - logout service


Dananji:
### Description

When the auth service block in the Canvas defines a [logout service](https://iiif.io/api/auth/2.0/#logout-service) for the current resource, Ramp should provide a logout method.
This provides the ability for the user to switch to a different account or close their session on a public terminal as needed.

In the current auth flow, an overlay is used on top of the VideoJS player to provide the UX needed for the login auth flow. Once the user is logged in successfully this overlay is unmounted. Therefore, a new UX pattern need to be derived to display the logout service information to the user.

### Done Looks Like
- [x] Investigate and design a logout UI for the player
- [x] Implement the logout UI design
- [x] Setup reducer methods to clear out any auth token/status saved in the state
- [x] Manually test the resource is not available to the user once logged out
- [x] Add unit tests the new functionality
- [x] Document the IIIF auth flow UX/UI in the Wiki, since this completes the auth implementation

joncameron:
Can be broken into two phases: look at the design, mock up UI (1 point) and then reevaluate to see how much effort the actual implementation will be.

Dananji:
Options for design in Figma: https://manor-limb-00546787.figma.site/

Option A: adds a new control to the control-bar once the user is logged in. This will not be as straight-forward as the other design option, because the control-bar is hidden during playback and this new icon could get drowned among the existing icons. However, this would be a good choice for audio players, because the control-bar is always visible in audio players and this space is readily available in control-bar.
Option B: adds a persistent badge to the top-right hand side of the player once logged in. This UI choice could be tricky in an audio player as it has very little space available in the container.

Dananji:
Notes:
- Code both designs
- Use option A for audio players and option B for video players
- Test and evaluate the designs in the demo site

Dananji:
Wiki: https://github.com/samvera-labs/ramp/wiki/IIIF-Auth-2.0-support

PR #979, opened 2026-07-01
Logout UI for IIIF Auth

Dananji:
Related issue: #958

Changes in this PR:
- add a logout UI,
  - for audio players: a `VideoJSAuthMenu` custom VideoJS `MenuButton` component. This is added to the control-bar before the `fullscreen` toggle when authenticated, showing a user icon that opens a popup with an optional label and a 'Log out' action button.
  - for video players: a persistent authenticated badge as a button with a popup menu with an optional label and a 'Log out' action button to the `AuthOverlay` component. This is shown once the user is authorized.
- both 'Log out' actions take the user to a new tab that presents the results of the HTTP GET request on the logout service’s URI, and clears the auth token and status in Context providers state
- once the user is logged out the player is cleaned up and disabled, so that authenticated media is removed from the player instance. And player aspect-ratio is reset to `16:9` to accommodate the display of the login modal in `AuthOverlay` component.

Additionally;
- move label parsing for all auth-related services into the parser function in `iiif-parser`, to cleanup the imports across UI components
- add relevant DOM-selectors to the `playerHotKeys()` in `utility-helpers.js` to enable keyboard accessibility for the new UI

Screenshot 2026-07-01 at 4 27 58 PM
Screenshot 2026-07-01 at 4 28 22 PM

### DIFF
--- a/src/components/MediaPlayer/VideoJS/components/js/VideoJSAuthMenu.js
+++ b/src/components/MediaPlayer/VideoJS/components/js/VideoJSAuthMenu.js
@@ -76,6 +76,7 @@ class VideoJSAuthMenu extends MenuButton {
     }
 
     const logOutBtn = new MenuItem(player_, { label: 'Log out', selectable: false });
+    logOutBtn.controlText('Log out');
     logOutBtn.el().insertAdjacentHTML('afterbegin', `
       <svg class="vjs-logout-icon" role="presentation">
         <use xlink:href="#log-out"></use>
--- a/src/components/MediaPlayer/VideoJS/components/js/VideoJSAuthMenu.js
+++ b/src/components/MediaPlayer/VideoJS/components/js/VideoJSAuthMenu.js
@@ -54,8 +54,8 @@ const MenuItem = videojs.getComponent('MenuItem');
 class VideoJSAuthMenu extends MenuButton {
   constructor(player, options) {
     super(player, options);
-    this.addClass('vjs-play-control vjs-control vjs-logout-menu');
-    this.setAttribute('data-testid', 'videojs-auth-icon');
+    this.addClass('vjs-play-control vjs-control vjs-auth-menu');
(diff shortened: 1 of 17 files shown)


src/components/MediaPlayer/VideoJS/components/js/VideoJSAuthMenu.js
```diff
@@ -0,0 +1,97 @@
+import videojs from 'video.js';
+import '../styles/VideoJSAuthMenu.scss';
+
+const userSVGSymbol = `
```
cjcolvar, line 4: Do these svg icons need to be defined here or can the ones in `services/svg-icons.js` be imported and used?

cjcolvar (approved):
This looks good as far as I can tell and it worked for me locally!  I had one small cleanup question that doesn't need to block merging.  Finally I do wonder if it would be good to allow hiding of the auth overlay badge along with the player controls so it doesn't have to cover the playing video if there isn't any user activity but we can test that out after merging.

src/components/MediaPlayer/VideoJS/components/js/VideoJSAuthMenu.js
Dananji: Good catch!! Yes, the defined `userSVGSumbol` and `logOutSVGSymbol` are exact duplicates of the `UserIcon` and `SignOutIcon` in `svg-icons.js` respectively.

This custom VideoJS component extends VideoJS's own base classes, thus mounting them using VideoJS's own explicit DOM lifecycle workflow which is completely outside of React's rendering model. The same symbol pattern is followed in `VideOJSTrackScrubber.js`.

However, I think we can remove this duplication by introducing some sort of a bridge functionality to turn the SVG icons from `svg-icons.js` into raw string usable in the VideoJS custom components following the existing symbol pattern. This would make the code more clean and maintainable in the future.

PR #981, opened 2026-07-02
Refactor playerHotKeys() to be generic

Dananji:
This PR is to refactor the `playerHotKeys()`function in `utility-helpers.js` to use a set of generic selectors to identify active interactive elements outside the player.
Previously the function maintained a hard-coded list of specific classNames and ARIA roles to determine when the player hotkeys should be suppressed.
To ensure, the existing functionality is preserved after the refactor without doing manual testing, I added unit-tests for the function and followed a TDD approach.

With this changes, hotkeys are now suppressed based on the *type* of an active element rather than specific component classNames, so new interactive UI components are covered automatically without needing updates to this function.

### DIFF
--- a/src/services/utility-helpers.js
+++ b/src/services/utility-helpers.js
@@ -712,7 +712,15 @@ export function playerHotKeys(event, player, canvasIsEmpty = false) {
   let playerInst = player?.player();
   let output = '';
 
-  let inputs = ['input', 'textarea', 'select'];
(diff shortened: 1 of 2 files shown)


cjcolvar (commented):
Looks great!  That's an amazing cleanup and nice set of tests!

There is a conflict blocking merging.  I can do another review when that is resolved.

cjcolvar (approved):

Issue #978, opened 2026-06-24
[BUG] IIIF Auth showing up in Avalon context


Dananji:
### Description

When Ramp is used in Avalon with the IIIF auth flow implementation, the auth pop up shows up for the restricted access media objects. This happens even when the user is already signed into Avalon. Ramp needs to handle this scenario in its auth flow.

<img width="957" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMGxRRGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b801b8fc809a7e944d70f59e55061840ec1a3e78/auth%20in%20avalon.png" alt="auth in avalon.png" />

### Done Looks Like
- [ ] Ramp identifies the auth status in the context it is used
- [ ] Depending on the identified auth status, display/hide the auth popup

PR #986, opened 2026-07-13
Fix auth

Dananji:
Related issue: #978

Changes in this PR:
- Add a new `auth.status` value `"authorized-external"` to hide logout menu widget/icon from the player when authentication is handled by an external app
- Once the resource is authenticated and the player is ready use VideoJS's `xhr-hooks-ready` event to attach `Authorization: Bearer <auth token>` to VideoJS's VHS requests for media resource fetch requests
- Bug fix: rename the `tokenService.heading` and `tokenService.note` to  `tokenService.errorHeading` and `tokenService.errorNote` which are the expected prop names in the components

### DIFF
--- a/demo/app.scss
+++ b/demo/app.scss
@@ -96,6 +96,10 @@ html {
     display: grid;
     gap: 0.1rem;
     grid-template-columns: repeat(auto-fit, minmax(240px, 1fr));
+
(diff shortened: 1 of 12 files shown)


cjcolvar (approved):
This looks good as far as I can tell!

Issue #983, opened 2026-07-06
Hide auth badge in video players on user-inactivity


Dananji:
### Description

When a user is authenticated to view restricted media using IIIF Auth 2.0, `AuthOverlay.js` displays a persistent "Authenticated" badge with a logout menu in the top-right corner of video players. This badge is always visible once authenticated. As a result, the badge permanently covers a corner of the video during playback, even after the rest of the player controls have faded out due to inactivity.

Relevant [comment](https://github.com/samvera-labs/ramp/pull/979#pullrequestreview-4639755057) from PR:
> I had one small cleanup question that doesn't need to block merging. Finally I do wonder if it would be good to allow hiding of the auth overlay badge along with the player controls so it doesn't have to cover the playing video if there isn't any user activity but we can test that out after merging.

### Done Looks Like
- [ ] Investigate how VideoJS signals control-bar inactivity and how if can be wired to the auth badge
- [ ] Hide the auth badge when player controls are hidden due to inactivity
- [ ] Show the auth badge when player controls reappear
- [ ] Ensure the logout menu closes appropriately if it's open when the badge is hidden
- [ ] Verify keyboard accessibility following the same pattern as VideoJS player controls

PR #987, opened 2026-07-14
Show/hide auth badge with video player controls during playback

Dananji:
Related issue: #983 

Use VideoJS's `'useractive'` and `'userinactive'` events to track the user (in)activity in the player instance. Using a custom hook to keep track of this in `ramp-hooks.js` avoids unnecessary re-renders to the `VideoJSPlayer` component tree and scopes the updates to `AuthOverlay` component.
Attaching VideoJS's `reportUserActivity()` function in the auth badge button's `onFocus` event handler triggers the `'useractive'` event to show player controls and auth badge when user tabs into the player from the top (because the first focus-able element inside the player is the auth badge from this direction).


cjcolvar (approved):
Looks great and really clean!  Thanks for coming back to this!

PR #988, opened 2026-07-14
Update circleCI to use node 26, add engines to package.json

Dananji:
Changes in this PR:
- add Node 26 to the `CircleCI` test matrix and drop the Node 20
- add an `engines` field to `package.json` with supported Node versions as `^20.19 || ^22.12 || >=24`

Notes: the `engines` field in `package.json` only limits the Node version used in the build/test environments of the library.


cjcolvar (approved):
Nice!

Issue #980, opened 2026-07-02
Investigate Libraries for Presentation 4 Manifest Support


joncameron:
## Description
The IIIF parsing library in Ramp will need to be updated. It doesn't appear that the parsing library used has been updated to work with the Presentation 4 spec, but we can use an existing library for support for these manifests.

## Investigation/Possible Options
- _manifesto_ library updated for v4 support (is this upcoming/extant?)
- other parsing libraries investigated (are there others with better support right now?)
- _parser_ library in iiif-commons has a draft PR for Prezi 4 support

## Done Looks Like
- [ ] Update the parsing library used in Ramp and parse info into Ramp data structures (Timeboxed)
- [ ] Create another ticket to explore options if using the current library is not tenable

Dananji:
IIIF-Commons parser library comparison

| Library | Description | Supported versions | Notes |
| --- | --- | --- | --- |
| `manifesto.js` | A IIIF Presentation API client and server utility library, that provides both a parser and a tree of class instances with getter fuctions for various resources in a IIIF Manifest. It is both a parser and an object model, which the `structures` parsing in Ramp relies on.  | Presentation 2 and 3 | Current feature development work is around extending support for 3D in the library while maintaining support for Presentation 2 and 3 with bug fixes. No mention of adding support for Presentation 4. |
| `@iiif/parser` | A more low-level parsing library, which converts the IIIF JSON-LD into a plain flat structured data. It seems that, the features of this library are focused on making it easier to develop with IIIF data extracted from a Manifest. | Presentation 2 and 3 and IIIF Image 3 (not-needed for Ramp) | Current [work](https://github.com/IIIF/3d/blob/main/temp-draft-4.md) is around adding support for Presentation 4 and 3D. |
| `@iiif/helpers` | A TypeScript utility library for building IIIF applications, where it acts as a query layer that can be built on top the data structure created by `@iiif/parser`. This query layer uses an in-memory reactive store created using Zustand to maintain the manifest data. This is especially helpful when, a IIIF client allows edits to the Manifest where it needs to refresh the Manifest data. For Ramp, this means in the current state it would need to maintain the same data in 2 different stores in-memory. | Presentation 2 and 3 | `@iiif/parser` is a peer dependency of this and should be installed alongside. There's a draft PR in the repo for in-progress Presentation 4 work, even though there's no mention of it in the README unlike `@iiif/parser`. |

[...]

PR #989, opened 2026-07-17
Parse presentation 4

Dananji:
Related issue: #980 

This PR adds support for parsing IIIF Presentation API v4 manifests with `manifesto.js`.

Changes in this PR:
- Add a new module named `iiif-version-parser.js` to do the following,
  - identify IIIF API version in a given Manifest. Parses the `@context` property in a given Manifest and defaults to 3 if the given context is not identified as 4
  - convert any given value for `motivation` property into an array, which is the accepted format for this property in v4. This was partially in-place for annotations parsing in Ramp
  - derive the placeholder IIIF resource property name based on the version. v3 --> `placeholderCanvas` and v4 --> `placeholderContainer`
- Add `type="Timeline"` support for parsing `items` in a given Manifest
- Add a condition to check the presence of `duration` property of a given Annotation with `type="Canvas"` to identify video resources against image resources
- Replace the use of `PropertyValue` class from `manifesto.js` with Ramp's function `getLabelValue()` which gives the same result

### DIFF
--- a/src/services/iiif-version-parser.js
+++ b/src/services/iiif-version-parser.js
@@ -1,0 +1,50 @@
+const PRESENTATION_4_CONTEXT = 'http://iiif.io/api/presentation/4/context.json';
+const PLACEHOLDER_PROP = { '3': 'placeholderCanvas', '4': 'placeholderContainer' };
+
+/**
+ * Determine the Presentation API version of a given Manifest from its
+ * '@context' property. Defaults to '3' when '@context' is missing or
+ * unrecognized. This is based on the fact that, Ramp initially only
+ * supported v3 without restrictions.
+ * @function VersionParser#getIIIFAPIVersion
+ * @param {Object} manifest IIIF Manifest
+ * @returns {String} IIIF Presentation API version ('3' or '4')
+ */
+export function getIIIFAPIVersion(manifest) {
+  const context = manifest?.['@context'];
+  const contexts = [].concat(context ?? []);
+  return contexts.includes(PRESENTATION_4_CONTEXT) ? '4' : '3';
+}
+
+/**
+ * Get the placeholder property name for a Canvas/Timeline or Annotation based on the
+ * IIIF Presentation API version of the given Manifest.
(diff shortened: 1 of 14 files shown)


src/services/iiif-version-parser.js
```diff
@@ -0,0 +1,50 @@
...
+ * Normalize an Annotation/AnnotationPage's 'motivation' to an array and if an expected
+ * motivation value is given check for its existence. Otherwise return the normalized
+ * motivation vlaues. This allows to read and parse 'motivation' values for both Presentations
+ * v3 and v4;
+ * - in Presentation v3 'motivation' is conventionally a single String (e.g. 'painting')
+ * - in Presentation v4 'motivation' is required to be an Array of strings (e.g. ['painting'])
+ * @function VersionParser#normalizeMotivation
+ * @param {String|Array} motivation 'motivation' value read off an Annotation
+ * @param {String|Null} expected given value to be checked whether it is included
+ * @returns {Array}
```
cjcolvar, line 51:
```suggestion
 * @returns {Array|Boolean}
```

cjcolvar: I'm wondering if this function should get split up since it seems to have two concerns:
- test if motivation is present
- get list of motivations present

My concern is it seems that it could be easy to make a mistake using it the wrong way that might not raise an error.  For example, in `utility-helpers.js` it is being used in an if clause:
```
if (resolveMotivation(p.motivation, motivation)) {
```
If the second argument is dropped
```
if (resolveMotivation(p.motivation)) {
```
then the if clause will always return true because the function returns an array object.

Do you think it would be good to split it up?

cjcolvar (commented):
This looks great!  Amazing that it isn't very much to support both prezi 3 and 4! ✨

src/services/iiif-version-parser.js
Dananji: You are correct! This could lead the code to fail silently. I'll split this up into 2 separate functions 👍

cjcolvar (approved):
I like the new function names!

PR #990, opened 2026-07-20
Combobox with sample manifests for demo site and update/add sample manifests

Dananji:
Add a new input combobox component to the demo site to display working demo manifests in the Ramp repo that can be used to illustrate different use-cases for Ramp components.
This new component replaces the input field for `Manifest URL` field in the demo site in Netlify with a field that provides a dropdown with existing sample manifests in the repo and a text field. This allows the user to either select a sample Manifest or type a different Manifest URL to be used by Ramp components on the page.

NOTE: merge after the Presentation 4 parsing work PR because the sample manifests include the manifests added in that PR.


cjcolvar (approved):
Nice!

What do you think about having the page initialize without any manifest url?  I think this would make the samples more visible since you wouldn't have to clear out the box to see the other options.

Dananji:
Demo site loads with the dropdown expanded without a pre-selection of a sample Manifest:

https://github.com/user-attachments/assets/79224386-df97-4749-be68-46784976533a

cjcolvar (approved):
Nice!

Issue #982, opened 2026-07-06
Remove duplicated SVG definitions in VideoJS custom component


Dananji:
### Description

Ramp has two separate places where the same SVG icons are defined for auth menu as,
- React components (e.g. `UserIcon`, `SignOutIcon`) in `src/services/svg-icons.js` used by React-rendered logout UI in `AuthOverlay.js`
- raw SVG `<symbol>` strings injected via `innerHTML` in VideoJS custom component `VideoJSAuthMenu.js` (e.g. `userSVGSymbol`, `logOutSVGSymbol`)

Relevant comment in PR: https://github.com/samvera-labs/ramp/pull/979#discussion_r3532055381

This duplication makes the code maintainability harder in the future, requiring synced updates in 2 places that could lead slight changes in the appearance going forward.

We should investigate a shared way to define an icon once in `svg-icons.js` and consume it from both React components and VideoJS custom components without duplication

### Done Looks Like
- [ ] Implement a bridge function for sharing icon definitions between React and VideoJS custom components
- [ ] Remove/migrate existing raw SVG `<symbol>` strings into `svg-icons.js`
  - [ ] `VideoJSAuthMenu.js` - `userSVGSymbol` and `logOutSVGSymbol` to reuse `UserIcon` and `SignOutIcon` respectively
  - [ ] `VideoJSTrackScrubber.js` - move `zoomOutIconSVG` and `zoomInIconSVG` to matching SVG React icons inside `svg-icons.js`
- [ ] Verify the UI appears as expected

PR #991, opened 2026-07-20
Cleanup duplicated SVG symbols in VideoJS custom components

Dananji:
Related issue: #982 

Changes in this PR:
- Remove duplicated SVG definitions as symbols in VideoJS custom components
- Introduce a bridge function in `svg-icons.js` to convert the React SVG icon components to SVG `<symbol>` strings in VideoJS components

### DIFF
--- a/src/services/svg-icons.js
+++ b/src/services/svg-icons.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import { renderToStaticMarkup } from 'react-dom/server';
 
 /** SVG icons for the edit buttons in Annotations component */
 export const EditIcon = () => {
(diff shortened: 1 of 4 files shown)


cjcolvar (approved):
Not as simple as I first imagined but still looks good.  Thanks for doing this cleanup!

Dananji:
Changes in the last commit:
- fix the zoom in/out SVGs in `svg-icons.js` which was flipped in the original React SVG icon definitions
- replace the use of `renderToStaticMarkup` in bridge function with `createRoot` and `flushSync` as suggested by the [docs](https://react.dev/reference/react-dom/server/renderToStaticMarkup#caveats)
- mark `react-dom/client` external in `vite.config.js`'s rollup config and map to the same `ReactDOM` UMD global as `react-dom`, so it stays aligned with the host app's copy of React instead of bundling a second one

src/services/svg-icons.js
```diff
@@ -204,3 +206,24 @@ export const UserIcon = () => {
...
+ * @param {Object} props props to pass to IconComponent when rendering
+ * @returns {String} `<symbol id="${id}">...</symbol>` markup string
+ */
+export const svgIconToSymbol = (IconComponent, id, props = {}) => {
+  const container = document.createElement('div');
+  const root = createRoot(container);
+  flushSync(() => root.render(<IconComponent {...props} />));
+
+  const svgEl = container.firstElementChild;
+  const viewBox = svgEl.getAttribute('viewBox') || '0 0 24 24' || '0 0 20 20';
```
cjcolvar: Unless I'm reading this wrong it seems like the third option is reachable.  Should the second option be dropped in favor of the third now that both of the SVGs use `0 0 20 20`?

cjcolvar (commented):
Nice catch with the vite config to avoid the react bundling!

cjcolvar (approved):

Issue #993, opened 2026-07-23
Support multi-canvas ranges


jamesmisson:
The IIIF Presentation API allows for multiple canvas items to be contained under a range (see the cookbook example here: https://iiif.io/api/cookbook/recipe/0031-bound-multivolume/).

We have AV material that will have logical ranges spanning across physical media - for instance, a recording that starts on side A of a tape and ends on side B. The IIIF structure would look like this:

```
{
          "id": "…range/r2",
          "type": "Range",
          "items": [ {"…canvas/1#t=30,60"}, {"…canvas/2#t=0,30"} ]
      }
```

I've edited Lunchroom Manners to have a second canvas (with the same files), and made the final range include a fragment of the second canvas.

https://gist.githubusercontent.com/jamesmisson/b78161e17714153fcc80bdee76f4a8cf/raw/05fc21c036dea5627b75bcc0946dc5ca2f129e48/multi-canvas-range.json

Ramp displays the range, but navigation to that range by clicking it doesn't work.  Would it be possible to display these structures in Ramp's navigation panel?

joncameron:
👍

PR #1000, opened 2026-08-04
Cross canvas ranges

Dananji:
Related issue: #993 

Changes in this PR:
- Parse and create separate navigational entries (structure item links) for each Canvas part in a multi-part Range in the `structures`. So that, they are displayed independently in the `StructuredNavigation` component.
  - an icon at the end of each link is displayed to convey they are part of a single Range
  - parts of a single Range has the same numerical item index suffixed by a lower-case letter
- Prioritize identifying multiple parts in the timespan at the end of the media file in `"ended"` event over the regular Canvas increment. E.g. in a Manifest with 3 canvases, if the Range at the end of the Canvas in the first Canvas has a cross-Canvas reference to the third Canvas the player skips the second Canvas and loads the third Canvas.
- Cleanup: move `VideoJSPlayer` component's `getActiveSegment()` helper to `useVideoJSPlayer` custom hook

In the following image `11a` and `11b` are two parts of the same Range that refers to 2 different canvases in the Manifest;

multi-part-ranges


cjcolvar:
@Dananji It looks like there are some conflicts that need to be resolved.  I can review after that.

src/services/ramp-hooks.js
```diff
@@ -714,8 +715,72 @@ export const useVideoJSPlayer = ({
...
+          activeSegment = {
+            canvasIndex: canvasIndex, duration: String(duration), homepage: '', id: canvasId,
+            isCanvas: true, isClickable: false, isEmpty: false, isRoot: false, isTitle: false,
+            itemIndex: canvasIndex + 1, items: [], label: label, rangeId: '', summary: summary,
+            times: range ? { start: range.start, end: range.end } : { start: 0, end: duration }
+          };
+        }
+      }
+      // For playlists timespans and canvasIdex are mapped one-to-one
+      activeSegment = canvasSegments[canvasIndex];
```
cjcolvar: Does this line need a guard or be inside an else?  As is it seems like it would overwrite what is set on line 735 (if that line gets executed).  I see in the version in VideoJSPlayer that these lines are return statements so execution doesn't continue like in this version.

src/services/iiif-parser.js
```diff
@@ -582,8 +584,42 @@ export function getStructureRanges(manifest, canvasesInfo, isPlaylist = false) {
...
+        };
+      };
+
+      /* When a Range has multiple items (Canvas parts), create one item per part to display
+      in the navigation display independetly. Sibling Canvas parts under the same Range has
+      the same item index suffixed by lower-case letters (e.g. 10a, 10b) */
+      if (!isCanvas && canvases.length > 1) {
+        const groupIndex = ++subIndex;
+        const partItems = canvases
+          .map((c, i) => parseCanvasPart(c, groupIndex, `${groupIndex}${String.fromCharCode(97 + i)}`))
```
cjcolvar: Definitely an edge case and one that probably doesn't need to be handled now but if i > 26 then this will move beyond a-z.  If this edge case needs a fix I assume we'd wrap and add a character (e.g. i = 26 would be `aa`).

src/services/iiif-parser.js
```diff
@@ -644,7 +683,7 @@ export function getStructureRanges(manifest, canvasesInfo, isPlaylist = false) {
         isCanvas: isCanvas,
         itemIndex: isCanvas ? cIndex : subIndex,
         canvasIndex: cIndex,
-        items: range.getRanges()?.length > 0 ? range.getRanges().map(r => parseItem(r, rootNode)) : [],
+        items: range.getRanges()?.length > 0 ? range.getRanges().flatMap(r => parseItem(r, rootNode)) : [],
```
cjcolvar, line 700: Interesting!  I've never seen `flatMap` before.

src/services/ramp-hooks.js
```diff
@@ -714,8 +715,72 @@ export const useVideoJSPlayer = ({
     }
   };
 
+  /**
+   * Get the timespan that encapsulates the current time of the playhead from the timespans in the
+   * current Canvas.
+   * @param {Number} time playhead's current time
+   * @returns {Object|null} active segment or null if not found
+   */
+  const getActiveSegment = (time) => {
```
cjcolvar, line 725: Does this method need tests or is it well covered by tests in other places?  VideoJSPlayer tests possibly?

cjcolvar (commented):
I think I'm a bit confused about what the expected modeling and behavior should be for manifests with non-linear playback of canvases listed in the items of a manifest.  I keep thinking that the structures/ranges should be like a playlist of explicit timespans and not an indexed link tree kind of like the transcripts. This is different than how ramp works.  When testing the new cross-canvas-range manifest I ran into this misconception again.  When range 11a plays into 11b ramp moves to the next canvas and the new canvas will keep playing until it finishes.  This makes sense with how ramp works but if I put a range (12) after that which references canvas 1 again then ramp will not continue back to canvas 1 after it finishes playing canvas 3.  Another example, if I set range 11b to start 10s into canvas 2 (instead of 0) ramp will start playback at 0 on canvas 3.  (A related confusion is this canvas has canvas 2 which isn't referenced in the structures and it's existence isn't displayed by ramp except in the duration in the Table of Contents root node of the structure.)

Maybe we could chat and you could help clear up my confusion around expected behavior of manifest structures/ranges?  I'm probably overthinking things and real world use cases aren't so complex.

Dananji:
> I think I'm a bit confused about what the expected modeling and behavior should be for manifests with non-linear playback of canvases listed in the items of a manifest. I keep thinking that the structures/ranges should be like a playlist of explicit timespans and not an indexed link tree kind of like the transcripts. This is different than how ramp works. When testing the new cross-canvas-range manifest I ran into this misconception again. When range 11a plays into 11b ramp moves to the next canvas and the new canvas will keep playing until it finishes. This makes sense with how ramp works but if I put a range (12) after that which references canvas 1 again then ramp will not continue back to canvas 1 after it finishes playing canvas 3. Another example, if I set range 11b to start 10s into canvas 2 (instead of 0) ramp will start playback at 0 on canvas 3. (A related confusion is this canvas has canvas 2 which isn't referenced in the structures and it's existence isn't displayed by ramp except in the duration in the Table of Contents root node of the structure.)
> 
> Maybe we could chat and you could help clear up my confusion around expected behavior of manifest structures/ranges? I'm probably overthinking things and real world use cases aren't so complex.

I was confused about this too. In the current implementation I was aiming to identify the Canvas parts in the Range at the end and then use that information for the Canvas switch. Therefore, it only respects timespans that can happen in the player's future.
During the implementation, I experimented with cross-Canvas Range items in arbitrary places in the structures (e.g. in the middle). But, didn't pursue those cases because BL's concern as far as I understood, was stitching 2 canvases via multiple Canvas parts in a single Range.

#### How Ramp interpret and utilize the `structures`;
In the current implementation, Ramp's `StructuredNavigation` component's behavior is simply displaying the hierarchical structure which is only a visual representation at the moment. For the playback experience Ramp is sticking with the default Canvas order in the Manifest and do not provide a way to utilize the alternate order set in the `structures` at the moment. 
Therefore, Ramp plays through the Canvases in the Manifest in their default order and highlights any Ranges it encounters on the way.

In other words, Ramp only provides the visualization of `structures` property without providing its functionality.

In order to fully utilize a Manifest's `structures`, Ramp should define a way to utilize the hierarchical order defined in `structures` which is different from the default ordering of Canvases in the Manifest. This can be recognized as an alternative playback mode?

Yes, it would be great to chat about this. And we can figure out the scope of BL's specific use-case for this work, discuss how Ramp can accurately interpret the `structures` and make that functional in Ramp's components?

src/services/ramp-hooks.js
Dananji: Yes it does! I'll add a couple in to `ramp-hooks.test.js`

src/services/ramp-hooks.js
Dananji: Good catch! I misunderstood the early returns in the previous version which would not reach the default case.

Dananji: I reverted this to the original flow, where the function returns early as it encounters each case without maintaining a variable for `activeSegment` to return at the end.

src/services/iiif-parser.js
```diff
@@ -582,8 +584,50 @@ export function getStructureRanges(manifest, canvasesInfo, isPlaylist = false) {
...
+      in the navigation display independetly. Sibling Canvas parts under the same Range has
+      the same item index suffixed by lower-case letters (e.g. 10a, 10b) */
+      if (!isCanvas && canvases.length > 1) {
+        const groupIndex = ++subIndex;
+        const partItems = canvases
+          .map((c, i) => {
+            // Build suffixes for i > 26 by stacking characters
+            let suffix = '';
+            while (i >= 0) {
+              suffix += String.fromCharCode(97 + (i % 26));
```
cjcolvar: I think this would make more sense to be prefixed.  e.g. 'ba' = 52 and 'ab' = 27.

```suggestion
              suffix = String.fromCharCode(97 + (i % 26)) + suffix;
```

cjcolvar (commented):
Looks good!  I had one small suggested change.

cjcolvar (approved):

PR #1002, opened 2026-08-04
Add a description to each Manifest option in demo combobox

Dananji:
Add a descriptive text to each sample Manifest option in the demo site's dropdown.


cjcolvar (approved):
This is a nice little improvement. 🙂

Issue #1001, opened 2026-08-04
[Regression] Transcript search only shows results from first matching annotation


cjcolvar:
**Describe the bug**
Transcript searches are only showing results from the first matching annotation.

**To Reproduce**
1. Load https://mco-staging.dlib.indiana.edu/media_objects/qn59r2417/manifest.json in ramp
2. Click on Transcripts tab
3. Search for "Unknown"
4. See "1 of 1 results"
5. Notice transcript selection dropdown shows more results in its label: "(48)"
6. Search for "not"
7. See "1 of 2 results"
8. Notice transcript selection dropdown shows more results in its label: "(12)"
9. Search for "good"
10. See "1 of 1 results"
11. Notice transcript selection dropdown shows more results in its label: "(10)"

**Expected behavior**
Ramp will show all relevant results returned by the search across all annotations.

**Screenshots**

<img width="856" height="548" alt="Image" src="https://github.com/user-attachments/assets/9520b12d-4b3d-4f74-8341-4b949dfde700" />

**Additional context**
This works on MCO so this appears to be a regression from ramp v4.0.2.

joncameron:
Working as expected on dev now 👍

PR #1003, opened 2026-08-05
Fix transcript content search for untimed text

Dananji:
Related issue: #1001 

When the text transcript has empty lines, the search response parser for untimed text offsets the entire search on the first pass. This skips the search parser from applying the necessary highlights to the transcript text, which is used for search hit navigation.

Once this PR is merged, I will cherry-pick to a branch off of the `v5.1.0` tag and do a patch release for `@samvera/ramp`.


cjcolvar (approved):
Fixes it! 🎉

Issue #998, opened 2026-07-29
Implement waveform display


Dananji:
### Description

This is the implementation work ticket related to #992

Create a custom VideoJS component similar to track-scrubber component in Ramp to display waveform data associated with the current Canvas. With the waveform-data library, we can use either the JSON or DAT file. Another possibility is the use of an image as representation of the waveform data. We'll try to support all three formats if possible, with image/PNG as lower priority than the others.

### Done Looks Like
- [x] Read and parse waveform data in the Canvas using BBC's `waveform-data` library  
- [x] Add a new prop to `MediaPlayer` component to enable waveform toggle control in player's control-bar
- [x] When a Canvas in Manifest doesn't have valid waveform information; the player control is hidden even when the waveform feature is turned ON
- [x] When valid waveform data is available with the feature turned ON in the `MediaPlayer` component;
  - [x] a player control icon is displayed in the player's control-bar 
  - [x] an audio player shows the waveform panel underneath the player by default and can be toggled using the player control
  - [x] a video player doesn't show the waveform panel by default and can be toggled using the player control

Notes:
- Obtain test manifests from BL to test for functionality and support for different formats they provide in their manifests for waveform

joncameron:
I tried to find examples from BL or others, but didn't come up with much.

Waveform rendering example (non-IIIF)
https://manawatuheritage.pncc.govt.nz/item/1a757eed-bb65-41ea-9e4b-32e1fd9444e9#resource_info

https://archive.org/details/hhfbc-cyl26 is mentioned in AV group docs as a test item that includes waveform json and images, at https://iiif.archive.org/iiif/hhfbc-cyl26/manifest.json.

In Clover:

<img width="1427" height="947" alt="Image" src="https://github.com/user-attachments/assets/e8036473-a466-450d-894d-6f169a7166ac" />

IA puts their waveforms in black and white PNG files at 800x200: https://ia800400.us.archive.org/9/items/hhfbc-cyl26/Band%201%20Clean.png

joncameron:
To note: this should work as expected with manifests containing many sources arranged as a single track or sequence. In this case there may be multiple waveforms to render. seeAlso is on the Canvas, so this could hopefuly be easier than it seems to make sure is working correctly.

PR #1007, opened 2026-08-13
Waveform display

Dananji:
Related issue: #998 

Changes in this PR:

- Add a new prop to `MediaPlayer` component named `showWaveform` to enable a toggle button to control visibility of a waveform display panel below the player (similar to track-scrubber)
- Read and parse waveform resources for a given Canvas in the following priority order;
  - a waveform dataset presented via `seeAlso` property (supports `.dat` and `.json` file types)
  - a waveform image presented via `accompanyingCanvas` property with a label that includes the text `waveform` (this guard against reading other images in a `accompanyingCanvas` as waveforms
  - the waveform information is read at the initial parsing stage for each Canvas, and then the actual waveform content is fetched as each Canvas is loaded into the player
- Create a new VideoJS custom component named `VideoJSWaveform` as a toggle button. This is wired to a static DOM element created as a panel identical to track-scrubber implementation, which displays the waveform visualization. 
- When the waveform control is enabled in `MediaPlayer` component props and the current Canvas has a valid waveform,
  - a waveform toggle control in the control-bar is displayed for both audio and video players
  - the waveform panel is expanded by default for audio players IF a saved playback position is not available for the Manifest
  - the waveform panel is always collapsed by default for video players
- Wire the waveform visualization using pointer event handlers to enable seek actions on the media playback
- Display the playback progress using the `background-image` CSS property as a tinted overlay on top of the waveform visualization. The area of this CSS property is clipped dynamically using the DOM property named `--range-progress` set by `VideoJSProgress` component, which aligns with the current playback progress

- Other helpful refactors:
  - move the saved playback position calculation logic into the `usePlaybackPosition` custom hook. This is used by both the waveform-display and resume-playback-modal implementations
  - reduce event handler dependent code in track-scrubber component identical to the new implementation in `VideoJSWaveform` 

audio-waveform

video-waveform

### Note: 
Still in progress work - adding support for multi-source canvases with a waveform presented using the `seeAlso` property for each Annotation body in the Canvas.

### DIFF
--- a/src/components/MediaPlayer/MediaPlayer.js
+++ b/src/components/MediaPlayer/MediaPlayer.js
@@ -23,9 +23,10 @@ const PLAYER_ID = 'iiif-media-player';
  * @param {Boolean} props.withCredentials
  * @param {String} props.language
  * @param {Object} props.resumeCache
- * @param {Boolean} props.resumeCache.enable whether to enable caching of playback positions
- * @param {Number} props.resumeCache.ttlDays number of days before a cached entry expires, default: 30 days
- * @param {Number} props.resumeCache.maxItems maximum number of entries to be stored in the cache, default: 200 entries
+ * @param {Boolean} props.resumeCache.enable
+ * @param {Number} props.resumeCache.ttlDays
+ * @param {Number} props.resumeCache.maxItems
+ * @param {Boolean} props.showWaveform
  */
 const MediaPlayer = ({
   enableFileDownload = false,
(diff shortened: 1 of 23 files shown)


src/components/MediaPlayer/MediaPlayer.js
```diff
@@ -23,9 +23,10 @@ const PLAYER_ID = 'iiif-media-player';
  * @param {Boolean} props.withCredentials
  * @param {String} props.language
  * @param {Object} props.resumeCache
- * @param {Boolean} props.resumeCache.enable whether to enable caching of playback positions
- * @param {Number} props.resumeCache.ttlDays number of days before a cached entry expires, default: 30 days
- * @param {Number} props.resumeCache.maxItems maximum number of entries to be stored in the cache, default: 200 entries
```
cjcolvar, line 28: Did these prop descriptions move somewhere else?

src/services/save-playback-positions.js
```diff
@@ -102,5 +106,26 @@ export const usePlaybackPositions = ({ enable, maxItems, ttlDays } = {}) => {
     setCache((current) => deleteCacheEntry(current, manifestURL));
   }, [enable, setCache]);
 
-  return { savePosition, getPosition, clearPosition };
+  /* Get the saved playback position for the current Manifest. Avoid reading saved position
+  from localStorage (using getPosition) if the current Manifest is a playlist Manifest or it
+  has a custom start position. */
+  const savedPosition = useMemo(() => {
```
cjcolvar, line 112: I like how this got refactored into here from VideoJSPlayer!

src/services/utility-helpers.js
```diff
@@ -1100,3 +1106,27 @@ export const sanitizeHTML = (html) => {
...
+ * 'format' property. Fallback to checking the resource extension against the MIME
+ * type to derive the data type and/or the format if 'format' is missing.
+ * @function Utils#getWaveformType
+ * @param {String} format MIME type of the waveform resource
+ * @param {String} id resource URL
+ * @returns {Object} { waveformType: 'data' | 'image' | null, format: String|null }
+ * resolved format falls back to the MIME type looked up from the resource's file extension
+ */
+export const getWaveformType = (format, id) => {
+  if (format === 'application/octet-stream') return { waveformType: 'data', format };
```
cjcolvar: Should json be tested here too?
```suggestion
  if (format === 'application/octet-stream' || format === 'application/json') return { waveformType: 'data', format };
```

cjcolvar (commented):
This is looking great as far as I can tell!

~How can I test this out locally?  I tried opening an MCO manifest with seeAlso waveform json files but I'm not seeing the panel toggle button.~ I tried it out with another MCO manifest and it works great!

src/components/MediaPlayer/MediaPlayer.js
Dananji: There is a more detailed prop description in the JSDoc block comment at the end of the file.

src/services/save-playback-positions.js
Dananji: I've been trying to move/refactor the helper functions in the `VideoJSPlayer.js` component because, that file is too long and is very hard to read and find relevant code :)
And this was a good opportunity to do that!

cjcolvar (approved):
Looks great!

Issue #999, opened 2026-07-30
[BUG] Better UX for invalid structure items in StructuredNavigation


Dananji:
### Description

Ramp displays timespans with both start and end times outside of the current Canvas's duration as regular structure items. i.e. they appears as links that can be clicked and perform the same functionality as the other timespan links in the `StructuredNavigation`.
E.g. last timespan named `"11.Putting Things Away (00:27)"` in the Presentation 3 - Lunchroom Manners Manifest in the [demo site](https://ramp.avalonmediasystem.org/).

For these invalid timespans Ramp needs to provide some kind of indication that, they are out of range and non-functional.

Suggestions:
- Making them plain text instead of links, and add a tooltip to indicate this is out of range
- Keep them as links, and add a tooltip to indicate this is out of range
- Keep them as links, and display some kind of a flash message when clicked

At the moment, Ramp doesn't have any tooltip or flash message infrastructure implemented. But building one doesn't seem too complex.

joncameron:
Keeping them as links might be an accessibility violation—DOM elements that are non-fuctional links probably aren't going to fly.

We can try out the first two options—either way, we can add a tooltip. 

@dananji will check the W3C design patterns to get a template for how to do this best while keeping accessibility compliance.

Dananji:
The solution uses the first option suggested in the ticket which is rendering the out-of-range timespans as plain-text `<span>` elements with an accessible tooltip.

PR #1013, opened 2026-08-26
Fix UX for out of range structure timespans

Dananji:
Related issue: #999 

Changes in this PR:
- Display out-of-range timespans as plain text with an accessible tooltip
- Use `tab-index` to include the plain text timespan to the page's tab-order so that, it can be included in the keyboard navigation within `StructuredNavigation` component using roving tabindex strategy
- Bug fix: move focus to the last interactive structure item in the `StructuredNavigation` component on `ArrowUp` key-press after initial focus on the component

Screenshot 2026-08-26 at 2 04 22 PM

**Note**: use the `'ramp--structured-nav__item-link'` class for the out-of-range timespan elements displayed using `<span>` HTML element with the addition of `'not-clickable'` class. This helps to keep their appearances consistent with the other timespans created with `<a>` HTML elements. Changing this class name at this point could break styling for current applications.


cjcolvar (approved):
Looks good!

PR #1014, opened 2026-08-28
Display range metadata in metadata display

jamesmisson:
This displays metadata on ranges in the `structures` property of a manifest. It updates to show the metadata of the current range, if it has any. I've used the metadata label as the heading in the display panel, but that's not set in stone - it might be better to leave it agnostic of content?

I've added some placeholder metadata to ranges in Lunchroom Manners to demo this: https://gist.githubusercontent.com/jamesmisson/ef50fe47bf2a8e34e2251c670858e769/raw/5c3d010d9c4cf2e6babab96f8d23ea7ae43648c2/range-metadata-demo.json

Each parsed structure item now carries a metadata field, populated by manifesto.js's `range.getMetadata()` method and then normalized to what Ramp expects. 

NB `getMetadata()` itself is unchanged and still only handles manifest and canvas metadata. Range metadata parsing happens in `getStructureRanges()`.

I've given `MetadataDisplay` a new prop, `displayOnlyRangeMetadata`, following the pattern from the other metadata sources.

I had to move `hasMetadata` from a `useState` to a computed value (l. 176) as adding a third source of metadata exposed a bug whereby the last source of metadata to set this state overrode the others (so if Ramp navigated to a canvas with no metadata, no metadata would be displayed even if there was some on the manifest level). I hope that's ok!


src/components/MetadataDisplay/MetadataDisplay.js
```diff
@@ -10,6 +10,7 @@ import cx from 'classnames';
  * related to the current resource. The display of the scope of this information
  * can be customized using props as needed.
  * @param {Object} props
+ * @param {Object} props.displayOnlyRangeMetadata
  * @param {Boolean} props.displayOnlyCanvasMetadata
```
Dananji, line 14: I think the type for this is meant to be `Boolean` to match its behavior and the type in `PropTypes` definition below?

src/components/MetadataDisplay/MetadataDisplay.js
```diff
@@ -53,11 +55,15 @@ const MetadataDisplay = ({
...
       // Display Canvas metadata only when specified in the props
-      const showCanvas = displayOnlyCanvasMetadata || displayAllMetadata;
+      const showCanvas = (displayOnlyCanvasMetadata || displayAllMetadata) && !displayOnlyRangeMetadata;
       setShowCanvasMetadata(showCanvas);
-      const showManifest = !displayOnlyCanvasMetadata || displayAllMetadata;
+      const showManifest = (!displayOnlyCanvasMetadata || displayAllMetadata) && !displayOnlyRangeMetadata;
       setShowManifestMetadata(showManifest);
 
+      // Display Range metadata only when specified in the props
+      const showRange = displayOnlyRangeMetadata || displayAllMetadata || !displayOnlyCanvasMetadata;
```
Dananji: This logic breaks the default behaviour for an existing implementation. For an example; an application that uses this component with its default props (like Avalon), this would enable showing any Range metadata by default.
However, the default behaviour is supposed to be displaying only the metadata at the Manifest-level.

I think changing this to `const showRange = displayOnlyRangeMetadata || displayAllMetadata` would restore the default behaviour?

src/components/MetadataDisplay/MetadataDisplay.js
```diff
@@ -160,6 +164,19 @@ const MetadataDisplay = ({
...
+  const rangeMetadataBlock = useMemo(() => {
+    if (showRangeMetadata && currentNavItem?.metadata?.length > 0) {
+      return (<>
+        <span>{currentNavItem.label}</span>
+        {buildMetadata(currentNavItem.metadata)}
+      </>);
+    }
+  }, [currentNavItem, showRangeMetadata]);
+
+  const hasMetadata = manifestMetadata?.length > 0
```
Dananji: > I had to move hasMetadata from a useState to a computed value (l. 176) as adding a third source of metadata exposed a bug whereby the last source of metadata to set this state overrode the others (so if Ramp navigated to a canvas with no metadata, no metadata would be displayed even if there was some on the manifest level). I hope that's ok!

This makes total sense, and it's a great catch! Moving `hasMetadata` to a computed value is definitely the right approach. I really appreciate you tracking down and fixing that bug.🚀

Dananji (changes requested):
Thanks for your contribution! 🎉 The changes look great overall.

Before we merge, I noticed a regression to the default behaviour of the component (please see my specific comment below regarding this)

Additionally, I have two quick clean-up requests for the repository setup:
- Could you please remove `.yarn/install-state.gz` from the PR and add it to the `.gitignore` file? It is a temporary local file generated by Yarn that shouldn't be tracked in Git, so adding it to `.gitignore` will help us prevent this in the future.
- I'm not quite ready to commit to a specific Yarn configuration change (`.yarnrc.yml`) for the repository just yet, as it needs an overall package manager setup first :) And I'd like to keep this PR focused strictly on the core feature changes. So, please remove this file from the PR as well.

Once those adjustments are made, I would be happy to review again and get this merged. Thank you!

src/components/MetadataDisplay/MetadataDisplay.js
jamesmisson: Thanks for spotting this, this was deliberate because I thought that you could just not put metadata on the range if you didn't want to display it, but I see that that is inconsistent with how canvas metadata works.

jamesmisson:
Hi Dananji, thanks for checking this so thoroughly and quickly! I've addressed those points in my recent commits.

Dananji (approved):
Looks great 🚀 Thanks for taking the time to make these changes!!

Issue #1008, opened 2026-08-14
Support for multi-source waveforms in a single Canvas


Dananji:
### Description

Extension of the work done in #998

Focuses on extending the waveform display work to support for multi-source canvases. A waveform can be presented for each source in the AnnotationBody in a single Canvas using the `seeAlso` property.
Ramp should be able to provide a contiguous presentation of waveform and media files in the entire Canvas upon reading and parsing a multi-source Canvas.

### Done Looks Like
- [x] Read and parse waveform and source information from multiple sources in a single Canvas
- [x] Display both waveform and media files for all sources in the Canvas in a single player instance
- [x] Provide continuous playback in both progress-rail and waveform
- [x] Able to drag/click to seek the waveform to advance the player time and vice-versa

PR #1015, opened 2026-08-28
Support display of waveform for each Annotation in multi-src Canvases

Dananji:
Related issue: #1008 

Changes in this PR:
- Parse waveform resources at each Annotation level in a Canvas for multi-source Canvases
- Combine dataset waveforms at each Annotation and display a contiguous waveform for the entire Canvas in the waveform panel
- Bug fixes;
  - Fix track-scrubber's handling of multi-source Canvases. The relevant use-case: when a timespan finishes playing while the track-scrubber panel is open it displays the entire Canvas duration in the track-scrubber. The track-scrubber needs to convey the source changes to VideoJS instance when the user seeks past the boundaries of the current source in the track-scrubber's progress-bar.

### DIFF
--- a/src/components/MediaPlayer/VideoJS/components/js/VideoJSTrackScrubber.js
+++ b/src/components/MediaPlayer/VideoJS/components/js/VideoJSTrackScrubber.js
@@ -302,7 +302,7 @@ class VideoJSTrackScrubber extends Button {
 
     /* Add the currentTrack's time (start time) to the track offset, as this is the Canvas
     relative time for the trackoffset derived from the pointer event. */
-    const offsetTime = trackoffset + currentTrackRef.current?.time ?? 0;
+    const offsetTime = trackoffset + (currentTrackRef.current?.time ?? 0);
 
     // Find the source corresponding to the pointer click event
     const clickedTarget = targets?.length > 1
--- a/src/components/MediaPlayer/VideoJS/components/js/VideoJSTrackScrubber.js
+++ b/src/components/MediaPlayer/VideoJS/components/js/VideoJSTrackScrubber.js
@@ -2,7 +2,7 @@ import { createRef } from 'react';
 import videojs from 'video.js';
 import '../styles/VideoJSTrackScrubber.scss';
 import '../styles/VideoJSProgress.scss';
-import { timeToHHmmss } from '@Services/utility-helpers';
+import { roundToPrecision, timeToHHmmss } from '@Services/utility-helpers';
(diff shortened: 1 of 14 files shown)


src/components/MediaPlayer/VideoJS/components/js/VideoJSTrackScrubber.js
```diff
@@ -284,16 +285,39 @@ class VideoJSTrackScrubber extends Button {
...
     }
+
+    const { altStart, srcIndex, targets } = player;
+
     let trackoffset = this.getTrackTime(e);
 
-    if (trackoffset != undefined) {
+    /* Add the currentTrack's time (start time) to the track offset, as this is the Canvas
+    relative time for the trackoffset derived from the pointer event. */
+    const offsetTime = trackoffset + currentTrackRef.current?.time ?? 0;
```
cjcolvar: There is an esbuild warning for this line (`warning: The "??" operator here will always return the left operand`) that the following seems to fix.  I guess `+` has higher precedence than `??`.
```suggestion
    const offsetTime = trackoffset + (currentTrackRef.current?.time ?? 0);
```

cjcolvar (approved):
This looks good!

I wonder if in the future ramp might need to handle waveforms for multi-src canvases not in the seeAlso on the painting annotations.  Like an array of waveforms in the seeAlso on the canvas.  This doesn't need to get handled in this PR if there isn't a current need for it but thinking that it might possibly come up in a manifest passed to ramp.

Dananji:
> I wonder if in the future ramp might need to handle waveforms for multi-src canvases not in the seeAlso on the painting annotations. Like an array of waveforms in the seeAlso on the canvas. This doesn't need to get handled in this PR if there isn't a current need for it but thinking that it might possibly come up in a manifest passed to ramp.

That's a great point! And I actually considered that use-case as well. But I ended up not pursuing it because, it wasn't entirely clear to me how the IIIF spec would explicitly map multiple waveforms in the Canvas's `seeAlso` property to their respective `painting` Annotations.

Currently, Ramp relies on a straightforward one-to-one mapping; either using `seeAlso` at the Annotation level for multi-source Canvases or at the Canvas level for single-source Canvases. If a Manifest puts multiple waveforms at the Canvas level, the mapping becomes a bit ambiguous, and Ramp would need an extra layer of logic built into it to figure out which waveform belongs to which source.
I agree it makes sense to leave this out for this PR, but we should definitely keep an eye out in case we start seeing Manifests structured that way!

PR #1016, opened 2026-08-28
Fix playhead and duration updates for multi-source switch

Dananji:
Reported issues:

When a multi-source Canvas is loaded into the player and seeking slightly past the currently loaded source's boundary; the following incorrect behavior is observed;
- duration: the player's duration changes to the current source's duration
- playhead position: jumps to the end of the progress-bar which doesn't accurately depict the user interaction

Fixes in this PR:
- Update the player's duration to the Canvas duration (total duration of all Annotations) on the `durationchange` which is re-reported by VideoJS's VHS on source change for multi-source Canvases. Previously this event overwrote the Canvas duration set in Ramp on this source change
- Fix the playhead position by intervening VideoJS's native progress-control method `getPercent()` which calculates and updates the progress percentage with respect to the player's duration. Due to a slight delay in frames, this function uses the current source's duration instead of the Canvas duration (total of all sources in the Canvas) for this calculation. And this led to the playhead jumping towards the end of the progress-bar instead of staying-in its correct position (the user's mouse position) with respect to the Canvas duration

### DIFF
--- a/src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
+++ b/src/components/MediaPlayer/VideoJS/VideoJSPlayer.js
@@ -207,6 +207,15 @@ function VideoJSPlayer({
       });
     });
 
+    /* VHS can re-report an individual source's duration via a 'durationchange' event
+    after the initial play/seek events. This silently over-writes the Canvas duration
+    set by Ramp. This is especially required for multi-source Canvases. */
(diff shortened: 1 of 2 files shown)


cjcolvar (approved):
Great work!  This seems like it would have been really hard to figure out and fix!

Issue #1005, opened 2026-08-10
Build and release process enhancements


Dananji:
### Description

The published releases for `@samvera/ramp` are created in local environment of a developer machine, which makes the built code to be relied on the local installations of dependencies (node_modules) and possibly other environmental factors. This creates different compiled code at different points of time for the same code and dependencies in `package.json`.

At the same time, the dev dependencies specified in the `package.json` used for the build process such as `vite`, `babel-plugin-react-compiler`, etc. are not locked at the moment. They are defined with a caret (`^`) symbol, which allows automatic updates to minor and patch versions of a dependency, but strictly locking it to the major version. So, the installed library version in `node_modules` can be slightly different from the version that is seen on the surface at different times.

Both of these combined create different bundled code for the same release at different points of time, which makes it harder to track changes between versions of the released code. And this eventually makes debugging process complex.

Having a build workflow for the release process that is independent from a local environment with locked versions of build libraries would make the built code for each release more reproducible over time.

### Done Looks Like
- [x] Locked dev dependencies for the build libraries for each release cycle
- [x] Create a machine independent build workflow (GitHub workflow?)
- [x] Document the release process

joncameron:
👍

PR #1019, opened 2026-09-03
Publish and release

Dananji:
Related issue: #1005

Changes in this PR:
- Add a new Git workflow file to publish the library to NPM and create a Release in GitHub, which is triggered by the creation of a new tag
- Since the new workflow keeps the build files as Git artifacts instead of contributing them back to the repo, abandon tracking build files in `dist/`. This needed to be tracked before because, the build was created locally and then pushed to remote for each release. Removing any tracked build files in the Git repo helps to create the source snapshot for the Git tag without any out-dated build files in the future
- Add a new markdown file named `RELEASING.md` to document the new release process and copy of it to `src/docs` for the Storybook site
- Add `.nvmrc` to lock a specific Node version for the build environment
- Add a `scripts/update-changelog.sh` shell script to generate `CHANGELOG.md` section for the new release
- Code changes:
  - The `ubuntu-latest` runners in GitHub actions usually come with Yarn Classic (1.x) pre-installed. And the GitHub actions in Ramp ran with whatever came pre-installed in these runners but, these installations can change silently. Therefore, to preserve reproducibility of builds add `"packageManager": "yarn@1.22.22"` to `package.json`, which ensures that all builds are created equally with a specific Yarn version. In order to enforce the use of this in the GitHub actions, add `run: corepack enable` to the workflow files
  - Add `"version"` hook to `package.json` to run the shell script to prepend a new section to the `CHANGELOG.md` with the merged PR information since the last release
  - Add missing `v5.1.1` tag to `CHANGELOG.md` and update the version in `package.json`
  - Change `vite.config.js` to put the dynamically imported VideoJS language files into a `dist/lang` directory

With these changes the new release process looks like this,
1. Confirm the code to be released is on `main` branch
2. Everything else is ready to be released --> all documentation (excluding `CHANGELOG.md`) and testing is complete
3. Checkout `main` branch locally and pull a fresh copy of the code
4. Run `npm version <newversion> | major | minor | patch...> --no-git-tag-version`
5. Confirm version bump is reflected in the `package.json` and a new section is created at the top of the `CHANGELOG.md` file with today's date and new version number
6. Review the changes locally and edit as needed
7. Commit, review, and merge changes
8. Create and push Git tag that matches the version bump made earlier in the format `v<version>`
9. Check the GitHub action `publish-and-release` is executed in "Actions" tab
10. Check the build package is staged in NPM and a Git Release is created
11. If all looks good, approve the staged package to be published to live registry

I tested the release workflow in Git in a private repository to ensure the tag push triggers the GitHub action and executes it without any errors and create the Release in GitHub repo. The only portion that was not tested was the `npm stage publish`, which can be tested during the next release.


.nvmrc
```diff
@@ -0,0 +1 @@
+22.23.2
```
cjcolvar, line 1: How does this version lock work with the `engines` config added in https://github.com/samvera-labs/ramp/pull/988?  Is the NVM configured version specifying the runtime of the build process and the build itself is packaging in a way that works for nodejs 20.19+?
Should the NVM node version be bumped to 26.x?

RELEASING.md
```diff
@@ -0,0 +1,48 @@
+# Releasing @samvera/ramp
+
+This document describes how to cut a release of `@samvera/ramp` component library and publish it to NPM.
+
+## Prerequisites
+
```
cjcolvar, line 6: The `update-changelog.sh` script calls the github CLI so I think it should be added as a prerequisite since it isn't obvious elsewhere that it is required (and I didn't have it already installed 🙂 ).
```suggestion

- GitHub CLI installed and setup (https://cli.github.com/)
```

.github/workflows/publish-and-release.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v6
+      - name: Setup Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version-file: '.nvmrc'
+          registry-url: 'https://registry.npmjs.org'
+          package-manager-cache: false # Never use caching in release builds
+      - name: Ensure npm CLI supports "npm stage publish" and trusted publishing
+        run: npm install -g npm@latest # Staged publishing needs npm >=11.15.0
```
cjcolvar, line 27: Not needed as part of this PR but do you think it makes sense to specify npm versions in package.json like node?  I'm not sure if it would be more effort than it is worth to keep the versions updated especially how they cross with node versions.  I'm asking this because I tried to install `npm@latest` locally and got an error because I'm not using NVM and only had node 22.14.0 installed.  (Upgrading to 22.23.2 resolved the error.)

cjcolvar (commented):
This is looking great!  I love how it allows for removing the compiled assets and `dist` directory! 🚀

.nvmrc
Dananji: Yes, the `.nvmrc` pins the Node.js version of the CI environment to one exact version for reproducibility of the built code.
The `engines` config in the `package.json` documents the supported Node.js version ranges for the consumers of the published library.
So, they both serve the purpose of managing the Node.js version of the library, but they function in fundamentally different ways.

Node 22 for the `.nvmrc` version was chosen because, it is the current lowest supported Node.js version. From what I found online, a library built with Node 22 will have upwards compatibility for 24 and 26 suggesting that, this is the safest option for the time being. 

However, EOL for Node 22 is April 30, 2027 so, we probably should consider bumping this before that date. Since 26.x is a newer version it will require more investigation and testing in different production environments to figure out the best fit. I will create a new ticket for this work.

.github/workflows/publish-and-release.yml
Dananji: I agree! This could be worth considering. Pinning `npm` alongside Node does add a bit of maintenance work to keep them in sync like you mentioned.

It is worth noting that, this step only ever runs in CI. Since the `.nvmrc` pins the Node version, this command will use the latest `npm` for that pinned Node version. I realize that this could resolve to slightly different `npm` versions at different times, but I'm guessing the drift is minimal at this point.

And yes, let's keep it out of scope of this as you suggested.

src/docs/Releasing.mdx
```diff
@@ -0,0 +1,52 @@
+import { Meta } from '@storybook/addon-docs/blocks';
+
+<Meta title="Introduction/Releasing" />
+
+# Releasing @samvera/ramp
+
+This document describes how to cut a release of `@samvera/ramp` component library and publish it to NPM.
+
+## Prerequisites
+
```
cjcolvar, line 10: Can you please add the following line to be consistent with the Releasing.md?
```suggestion

- GitHub CLI installed and setup (https://cli.github.com/)
```

cjcolvar (approved):
Looks great!

src/docs/Releasing.mdx
Dananji: Yes! 🎉 I added a bit more information to the text in the next commit.

Issue #1009, opened 2026-08-14
Add support for parsing 'provides' property in IIIF v4


Dananji:
### Description

IIIF Presentation API 4.0 spec introduces a new property called [`provides`](https://iiif.io/api/presentation/4.0/model/#provides) for `supplementing` Annotations. This property informs of additional functionality and features related to a resource in the Annotation. This can be often used for accessibility purposes and gives a broader context of the usage of the linked resource that can be used by a IIIF client to present the resource in a more meaningful way to the user.

The value of a `provides` property is an array of strings, where each string identifies a particular feature and must be taken from the `provides` [registry](https://iiif.io/api/registry/accessibility/).

### Done Looks Like
- [x] Ramp reads and parses `provides` property in `supplementing` Annotations in a v4 Manifest
- [x] Map each of the values in the registry to existing component/UI in Ramp 
- [ ] Discuss/research the need for addition of new UI to display resources as needed that cannot be presented in Ramp's existing UI/components
- [ ] Create new tickets for any new UI implementations as needed

joncameron:
Need to make sure support for this is part of Presentation 4 implementation in Avalon.

Dananji:
Ramp maps a11y values in IIIF [registry of accessibility](https://iiif.io/api/registry/accessibility/) via `provides` property in `"supplementing"` Annotations to Ramp's internal data structure as follows;
  - transcript --> displayed in both Transcript and Annotations components
  - closedCaptions/subtitles --> displayed in player as captions
  - audioDescription --> displayed in player as AD track, and in both Transcript and Annotations components

When multiple values are given for `provides` property, assess each value and build internal data structure as needed for different UI components. e.g. when `provides=["transcript", "closedCaptions"]` create both a transcript annotation for Transcript component and a `<track>` element for the player

PR #1024, opened 2026-09-10
Integrate 'provides' prop values to decide display of Annotations

Dananji:
Related issue: #1009 

Changes in this PR:
- Read `provides` property from Annotations regardless of the IIIF version defaulting an empty array when the property does not exist because, even in IIIF Presentation v4 Manifests can have `"supplementing"` Annotations without `provides` property
- Map a11y values in IIIF [registry of accessibility](https://iiif.io/api/registry/accessibility/) via `provides` property in `"supplementing"` Annotations to Ramp's internal data structure as follows;
  - transcript --> displayed in both Transcript and Annotations components
  - closedCaptions/subtitles --> displayed in player as captions
  - audioDescription --> displayed in player as AD track, and in both Transcript and Annotations components
- When multiple values are given for `provides` assess each value and build internal data structure as needed for different UI components. e.g. when `provides=["transcript", "closedCaptions"]` create both a transcript annotation for Transcript component and a `<track>` element for the player
- Rename `normalizeMotivation()` in `iiif-version-parser.js` to `normalizeValues() to generalize its use-case for both `motivation` and `provides` values

### DIFF
--- a/public/manifests/dev/lunchroom-manners-v4.json
+++ b/public/manifests/dev/lunchroom-manners-v4.json
@@ -111,6 +111,7 @@
 								"supplementing"
 							],
 							"provides": [
+								"audioDescription",
 								"subtitles"
 							],
 							"body": {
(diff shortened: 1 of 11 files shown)


cjcolvar (approved):
Looks great!  It is nice to see new functionality come in for v4 support alongside the continuing support for v3.

